PR #114, opened 2022-01-14
WIP: Add Equinix Metal bootstrap

DavidSpek:
Work in progress, do not merge.


minio/helm/minio/values.yaml
```diff
@@ -5,112 +5,20 @@ secret:
 storageClass: ""
 
 minio:
-  ## Provide a name in place of minio for `app:` labels
```
michaeljguarino, line 8: I'm assume the red diffs on this file is just removing duplication of existing defaults?

postgres/helm/postgres/templates/secrets.yaml
```diff
@@ -1,11 +1,13 @@
+{{- if .Values.configSecret.enabled }}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: postgres-environment
+  name: plural-postgres-s3
```
michaeljguarino, line 5: we should probably keep this secret the same name, as moving it very likely will cause all the dbs we've deployed to restart.  The configmap change might do that as well, though.  This also means making it optional is probably a no-go.

michaeljguarino: we should test it on a non-prod cluster to ensure this doesn't happen at least on aws so we don't bring down app.plural.sh for a bit with it

minio/helm/minio/values.yaml
DavidSpek: Yeah, I did a major upgrade of the MinIO chart and removed all the default values since that causes a lot more maintenance overhead.

PR #115, opened 2022-01-25
Add default node groups and set labels/taints

DavidSpek:
This PR adds a bunch of default node groups with the necessary labels and taints for us to select them. The labels and taints are added as tags to the ASGs so they are compatible with the cluster autoscaler. Should be merged after https://github.com/pluralsh/terraform-aws-eks/pull/1.


michaeljguarino (approved):
tested locally, and looks good to go!

PR #153, opened 2022-02-10
AWS bootstrap: remove multi-AZ node groups

DavidSpek:
This PR will remove the old multi-AZ node groups that can cause problems with EBS volumes to fully transition over to single-AZ node groups.


bootstrap/terraform/aws-bootstrap/main.tf
```diff
@@ -78,17 +65,21 @@ module "single_az_node_groups" {
...
+  node_groups            = var.multi_az_node_groups
+  set_desired_size       = false
+  private_subnet_ids     = module.vpc.worker_private_subnets_ids
+
+  ng_depends_on = [
+    module.cluster.config_map_aws_auth
+  ]
+}
 
 resource "aws_eks_addon" "vpc_cni" {
```
michaeljguarino, line 84: we probably need to rewire the depends_on fields for the aws_eks_addons here to point to the new node group modules.

bootstrap/terraform/aws-bootstrap/output.tf
```diff
@@ -32,5 +32,5 @@ output "cluster_public_subnet_ids" {
 }
 
 output "node_groups" {
```
michaeljguarino, line 34: could we also grab the first groups worker_role_arn to make life easier for downstream apps?

PR #154, opened 2022-02-10
Remove pod security policies from monitoring charts

michaeljguarino:
Needed to support k8s 1.22


lycanlee:
Thanks for jumping on this so quickly @michaeljguarino!

PR #175, opened 2022-02-28
set k8s version explicitly in bootstrap

### LINKED ISSUE #165: Pin k8s versions in core bootstrapping terraform modules
## Use Case
We had a user pick up a gcp deploy and realize quickly lots of apps were broken due to GKE stealthily moving to k8s 1.22.  Our version infrastructure should provide us the tools to manage this, but we need to use it properly, in particular by pinning our `{provider}-bootstrap` modules to specific k8s versions.

## Ideas of Implementation
Go through the terraform and find the appropriate settings to use for each module.

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

ycryptx:
## Summary
This makes it so the k8s version across our cloud providers is explicitly defined.
closes #165 

## Checklist

### DIFF
--- a/bootstrap/terraform/aws-bootstrap/main.tf
+++ b/bootstrap/terraform/aws-bootstrap/main.tf
@@ -35,7 +35,7 @@ module "vpc" {
 module "cluster" {
   source          = "github.com/pluralsh/terraform-aws-eks?ref=always-create-auth-cm"
   cluster_name    = var.cluster_name
-  cluster_version = "1.21"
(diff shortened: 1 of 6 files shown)


DavidSpek (approved):
Looks good to me.

DavidSpek:
@ycryptx For GKE you should set it to the current latest version, otherwise it will cause a downgrade for people which isn’t supported.

PR #181, opened 2022-03-07
GKE: Pin K8s version, use cilium, setup similar node group to EKS

DavidSpek:
Requires: https://github.com/pluralsh/plural/pull/178

## Summary
Pins the GKE kubernetes version. Uses Cilium as the CNI. Sets up the same type of node groups as we use on EKS with the same labels. Also deploys the FileStore CSI driver.

## Checklist

### DIFF
--- a/bootstrap/terraform/gcp-bootstrap/main.tf
+++ b/bootstrap/terraform/gcp-bootstrap/main.tf
@@ -47,8 +47,8 @@ module "gke" {
   http_load_balancing        = true
   remove_default_node_pool   = true
   add_cluster_firewall_rules = true
-  network_policy             = false
-  datapath_provider          = "ADVANCED_DATAPATH"
+  network_policy             = var.network_policy_enabled
+  datapath_provider          = var.datapath_provider
   kubernetes_version         = var.k8s_version
   filestore_csi_driver       = true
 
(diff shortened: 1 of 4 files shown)


bootstrap/terraform/gcp-bootstrap/main.tf
```diff
@@ -40,10 +40,17 @@ module "gke" {
   http_load_balancing        = true
   remove_default_node_pool   = true
   add_cluster_firewall_rules = true
-  network_policy             = true
+  network_policy             = false
```
michaeljguarino, line 43: what's the implication of setting this to false?  Is it duplicative with cilium?

bootstrap/terraform/gcp-bootstrap/variables.tf
```diff
@@ -96,6 +239,11 @@ instances are launched.
 EOF
 }
 
+variable "k8s_version" {
+  type = string
+  default = "1.22.6-gke.300"
```
michaeljguarino, line 244: 💯

DavidSpek: should be bumped to `1.22.7-gke.900`

DavidSpek:
@michaeljguarino I've tested both an upgrade from the current latest version and a fresh deployment and everything went smoothly so I think this can be merged.

michaeljguarino (approved):
let's get this merged for sure

PR #184, opened 2022-03-16
GCP set bucket locations using new CLI

DavidSpek:
## Summary
This PR updates the terraform artifacts for GCP so that the bucket location is set using the new CLI changes from https://github.com/pluralsh/plural-cli/pull/30. 

## Checklist

### DIFF
--- a/bootstrap/terraform/gcp-bootstrap/locals.tf
+++ b/bootstrap/terraform/gcp-bootstrap/locals.tf
@@ -1,9 +1,6 @@
 locals {
-	gcp_location       = var.gcp_location
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-  pods_cidr_name     = "${var.gcp_location}-${var.cluster_name}-pods"
-  services_cidr_name = "${var.gcp_location}-${var.cluster_name}-services"
+  pods_cidr_name     = "${var.gcp_region}-${var.cluster_name}-pods"
+  services_cidr_name = "${var.gcp_region}-${var.cluster_name}-services"
   vpc_network_name   = var.vpc_network_name != "" ? var.vpc_network_name : "${var.vpc_name_prefix}-network"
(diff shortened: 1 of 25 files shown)


bootstrap/terraform/gcp-bootstrap/locals.tf
```diff
@@ -1,9 +1,6 @@
 locals {
-	gcp_location       = var.gcp_location
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-  pods_cidr_name     = "${var.gcp_location}-${var.cluster_name}-pods"
-  services_cidr_name = "${var.gcp_location}-${var.cluster_name}-services"
+  pods_cidr_name     = "${var.gcp_region}-${var.cluster_name}-pods"
+  services_cidr_name = "${var.gcp_region}-${var.cluster_name}-services"
```
DavidSpek, line 3: @michaeljguarino the cidrs are regional, so the location name doesn't make much sense. However, this change is likely breaking.

michaeljguarino: my guess is this could fall into a full control plane replacement, which is going to be very bad.

DavidSpek: That's avoided by the changes in https://github.com/pluralsh/plural-artifacts/pull/184/files#diff-03cedf9cd127cd1e99af028f7ba2e13316f5cc1b71a28e6e06f0e068e0b2220f.

michaeljguarino: ok we should verify but that looks good enough to solve for.

DavidSpek: I've verified that particular fix. Without it, planning bootstrap will actually fail because the kubernetes provider can't connect to the cluster (it tries to connect to localhost since terraform is planning to delete the existing cluster so there aren't any connection details).

michaeljguarino (approved):

PR #198, opened 2022-03-29
fix aks version variable reference

RickRen7575:
## Summary
This should fix the k8s version variable reference in terraform.

## Test Plan
I have ensured that the variable reference is matching across files.

## Checklist


DavidSpek:
Thanks for the contribution!

PR #220, opened 2022-04-14
Fix airflow azure logs

michaeljguarino:
## Summary
We needed to prefix the minio endpoint with https://.  Otherwise credentials are all properly placed and seems to function.

## Test Plan
`plural link helm airflow --name airflow --path ../plural-az-installations/airflow/helm/airflow` and kicked off a dag run to check its logs

## Checklist
- [x] My change requires a change to the documentation and I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.


RickRen7575:
Thanks for sharing this! I see the `plural link...` command now and how it works

PR #224, opened 2022-04-20
airflow helm updates

mertkayhan:
## Summary

Hello @michaeljguarino , we are currently using Airflow through Plural with the following modifications:
* We use Cloud SQL as our db
* We use Kubernetes Executor instead of Celery based one
* We have a custom GCS sync operator which utilizes existing PVC to sync dags

Based on these modifications we do not necessarily need Zalando Postgres, Redis, etc. instances. Therefore, I updated the helm templates to respect these variations.

I am new to Plural, so I would appreciate some feedback on this :) 

## Checklist

### DIFF
--- a/airflow/helm/airflow/templates/postgres.yaml
+++ b/airflow/helm/airflow/templates/postgres.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.postgres.enabled }}
 apiVersion: acid.zalan.do/v1
 kind: postgresql
 metadata:
@@ -85,3 +86,4 @@ spec:
     targetPort: http-metrics
   selector:
     spilo-role: replica
+{{- end }}
--- a/airflow/helm/airflow/values.yaml.tpl
+++ b/airflow/helm/airflow/values.yaml.tpl
@@ -4,7 +4,7 @@ global:
     - description: airbyte web ui
       url: {{ .Values.hostname }}
 
-{{ if .Values.redisEnabled }}
+{{ if not .Values.redisDisabled }}
 secrets:
   redis_password: {{ dedupe . "airflow.secrets.redis_password" (randAlphaNum 14) }}
 {{ end }}
@@ -15,6 +15,7 @@ s3access:
   secret_access_key: {{ importValue "Terraform" "secret_access_key" }}
 {{ end }}
 
+{{ if not .Values.gitSyncDisabled }}
 sshConfig:
 {{ if .Values.hostname }}
   id_rsa: {{ ternary .Values.private_key (dedupe . "airflow.sshConfig.id_rsa" "") (hasKey .Values "private_key") | quote }}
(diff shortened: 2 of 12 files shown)


DavidSpek:
@mertkayhan Thank you very much for the contribution. I’ll look at your changes tomorrow and leave a review for you.

mertkayhan:
Thanks @DavidSpek, much appreciated

michaeljguarino (approved):
This is amazing!  Thanks for all the work here.  We'll probably want to modify the airflow recipes after merge to not ask for dag git info if you chose to opt out of git sync.

One other question, are you guys using your own airflow docker image then?

michaeljguarino:
@DavidSpek we should do a test on one of our clusters to ensure it still works with local pg and redis setups, but I think it should be fine based on my spot-check of the code

mertkayhan:
@michaeljguarino No, we are using the upstream image. Initially, we injected a custom sidecar into the containers but now it is a separate deployment which shares an NFS with Airflow components.

airflow/helm/airflow/values.yaml.tpl
```diff
@@ -160,11 +164,13 @@ airflow:
 
   dags:
     gitSync:
-      enabled: true
+      enabled: {{ .Values.gitSyncEnabled | ternary "true" "false" }}
```
michaeljguarino, line 167: Ok I've tested this locally and just realized this setting will break existing airflow installs.  We'd probably need to set it to a default-falsy setting, like:

```
{{ if not .Values.gitSyncDisabled }}
  dags:
    gitSync:
       ....
{{ end }}
```

Since people won't have the gitSyncEnabled flag set in their context by default.

airflow/helm/airflow/values.yaml.tpl
```diff
@@ -4,8 +4,10 @@ global:
     - description: airbyte web ui
       url: {{ .Values.hostname }}
 
+{{ if .Values.redisEnabled }}
```
michaeljguarino, line 7: Similar issue here, use {{ if not .Values.redisDisabled }} to turn this off manually but default-on.

airflow/helm/airflow/values.yaml.tpl
```diff
@@ -22,7 +24,9 @@ sshConfig:
   id_rsa_pub: example
 {{ end }}
 
+{{ if .Values.postgresqlEnabled }}
```
michaeljguarino, line 27: like with redis and gitsync, this should default on, so we'd need something like:

{{ if not .Values.postgresDisabled }}

michaeljguarino (changes requested):

airflow/helm/airflow/values.yaml.tpl
mertkayhan: Good point, agreed

airflow/helm/airflow/templates/postgres.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.postgres.enabled }}
```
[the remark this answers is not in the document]
michaeljguarino: Also, it's kind of unfortunate, but the safest way to ensure this is to put it into the values.yaml.tpl, I've seen the internal values.yaml not consistently apply after helm installs (some internal helm issue, I suppose)

PR #225, opened 2022-04-25
add hasura admin secret to notes.tpl

michaeljguarino:
## Summary
We don't currently print the hasura admin secrets in the deployment notes, so users have to manually fetch it.

## Test Plan
can't test locally unfortunately

## Checklist


DavidSpek (approved):
Looks good to me

PR #240, opened 2022-04-29
airbyte helm updates

mertkayhan:
## Summary

Hello @michaeljguarino, similar to the Airflow case we are using an external db for our Airbyte deployment. I modified the helm templates to support that. I would appreciate if you could take a look.

## Checklist

### DIFF
--- a/airbyte/helm/airbyte/values.yaml
+++ b/airbyte/helm/airbyte/values.yaml
@@ -28,7 +28,7 @@ postgres:
         cpu: 100m
         memory: 100Mi
       limits:
-        cpu: '2'
+        cpu: "2"
         memory: 1Gi
 
 oidcProxy:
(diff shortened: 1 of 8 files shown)


michaeljguarino (changes requested):
Similar to the airflow deployment, we should make sure this defaults postgres to on.  Best strategy there is to set it in both values.yaml and values.yaml.tpl to ensure it takes despite helm's weirdness

airbyte/helm/airbyte/values.yaml
```diff
@@ -45,131 +45,152 @@ private:
...
+              "temporal",
+              "resources",
+              "requests",
+              "memory",
+            ]
 
 airbyte:
   postgresql:
-    enabled: false
+    enabled: true
```
michaeljguarino, line 179: this is a bit of a pain, but this actually has a side-effect of causing the bitnami postgres helm chart to also be deployed, which we'd definitely prefer to avoid in preference of zalando's pg operator.  Perhaps we could use a different helm value to toggle this, or move it to the top-level postgres field?

mertkayhan: ah fair point, I did not pay attention
I will fix that

michaeljguarino: yeah that looks like the right approach to me!

michaeljguarino:
Ok I'll try to test this locally, and if it's all good, we can get it merged.

michaeljguarino (approved):
Ok just tested locally and it seems like everythign is safe.  Good work on getting this built out!

PR #252, opened 2022-05-13
openvpn changes

fayaz-abdul:
## Summary

attempting to create artefact for Openvpn application

new feature 

## Checklist


openvpn/terraform/gcp/main.tf
```diff
@@ -0,0 +1,11 @@
+resource "kubernetes_namespace" "openvpn" {
```
michaeljguarino, line 1: Since gcp, aws, azure, etc really have no cloud interaction here, should we perhaps merge these into a single terraform module supporting all providers?

michaeljguarino (commented):
It looks like this is just locally copying the helm stable openvpn chart, is there any reason why we didn't use this which seems to be somewhat maintained: https://artifacthub.io/packages/helm/stenic/openvpn-as?

openvpn/terraform/gcp/main.tf
DavidSpek: Agreed

DavidSpek:
Personally, I think creating a helm chart from https://github.com/juanfont/headscale/tree/main/docs/examples/kustomize and then deploying a [subnet router](https://tailscale.com/kb/1019/subnets/) should be explored. That would allow us to integrate with OIDC and allow users to set ACLs for who can access what services.

DavidSpek:
More docs on Tailscale on Kubernetes. https://tailscale.com/kb/1185/kubernetes/

Except for a subnet router, adding tailscale sidecars to applications so they are added to the network directly might be a better approach as it has some benefits. We could do something similar to our OAuth injector for this.

PR #257, opened 2022-05-18
Onboard Vault

fabricetriboix:
Onboard Vault (still work in progress)


vault/terraform/aws/main.tf
```diff
@@ -0,0 +1,61 @@
...
+data "aws_eks_cluster" "plural" {
+  name = var.cluster_name
+}
+
+locals {
+  account_id = data.aws_caller_identity.current.account_id
+  oidc_provider = regex("^(https?://)(.*)$", data.aws_eks_cluster.plural.identity[0].oidc[0].issuer)[1]
+}
+
+module "assumable_role_vault" {
```
michaeljguarino, line 51: you can actually use: https://github.com/pluralsh/module-library/tree/main/terraform/irsa-policy to encapsulate this IRSA stuff

michaeljguarino (approved):
let's go ahead and merge this to get the resources synced, and then create a new branch for future work

fabricetriboix:
> I actually pushed the changes I did to Plural directly already so no need to merge it to get the resources synced.
> 
> @fabricetriboix why did you change the path from /oidc to /plural? I believe using a non-standard path for the integration will mean users need to manually input the path when trying to login, which will create additional friction. Our approach up to now has mostly been to use defaults and standards to reduce possible friction for end users.

There's a few reasons:
 1. The sample code provided to me by a user on the Vault forum was doing this
 2. I find it easier to reason with than a path that contains something like `/oidc/oidc`
 3. If a plural user wants to setup up their own OIDC, then are free to use the default path if they want to

The `vault auth tune -listing-visibility=unauth plural` command that I added makes the plural OIDC visible in the Vault UI, see [https://vault.ft5.onplural.sh/](https://vault.ft5.onplural.sh/).

If you believe that's a problem, I can revert back to the previous path.

PR #260, opened 2022-05-24
update hasura chart to plug in cloud sql db

mertkayhan:
## Summary

Hello @michaeljguarino, we are currently onboarding Hasura and would like to use Cloud SQL as the Postgres db. I did some changes to support that. Could you take a look?

## Test Plan

Local testing with plural link

## Checklist

### DIFF
--- a/hasura/helm/hasura/templates/postgres.yaml
+++ b/hasura/helm/hasura/templates/postgres.yaml
@@ -1,4 +1,4 @@
-{{- if eq .Values.hasura.postgresql.host "plural-hasura" }}
+{{- if .Values.hasura.postgresql.operatorEnabled }}
 apiVersion: acid.zalan.do/v1
 kind: postgresql
 metadata:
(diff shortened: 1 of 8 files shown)


hasura/helm/hasura/templates/postgres.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.hasura.postgresql.host "plural-hasura" }}
```
DavidSpek, line 1: Maybe it would be better to use another key for this. Something like:

```yaml
hasura:
  pgClient:
    external:
      enabled: true
      operatorEnabled: true
```

DavidSpek (commented):
Thanks for the contribution. Overall it looks good to me, just one small comment. Also, you'll need to bump the chart version before the PR can be merged.
@michaeljguarino I think you onboarded Hasura so maybe you should also look at this.

michaeljguarino (approved):
lgtm

PR #263, opened 2022-05-27
Add admin group and final cleanup

fabricetriboix:


vault/helm/vault/values.yaml.tpl
```diff
@@ -43,14 +45,6 @@ vault:
         eks.amazonaws.com/role-arn: "arn:aws:iam::{{ .Project }}:role/{{ .Cluster }}-vault"
     {{ end }}
 
-    {{ if .OIDC }}
```
michaeljguarino, line 46: Don't we still need these?

fabricetriboix: No, we don't, it's just dead code. It's something I started, but David changed how it's done. The client id and client secret are in a secret and mounted as environment variables (see a bit later in the file). The discovery URL is set in a regular environment variable (see a bit earlier in the file).

PR #277, opened 2022-06-02
Vault: fix vault policy + jobs + set owner to admin

DavidSpek:
## Summary
This PR fixes the Vault admin permissions and adds a script that ensures they are updated before executing following commands. It also changes the OIDC behavior so that the Plural user that deployed Vault gets bound to the admin policy when OIDC is enabled.

### DIFF
--- a/vault/helm/vault/templates/vault-configmap.yaml
+++ b/vault/helm/vault/templates/vault-configmap.yaml
@@ -6,38 +6,71 @@ metadata:
   {{ include "vault-plural.labels" . | nindent 4 }}
 data:
   admin-policy.hcl: |
-    # Allow everything
-    path "*" {
+    # Read system health check
+    path "sys/health"
+    {
+      capabilities = ["read", "sudo"]
+    }
+
+    # Create and manage identities.
+    path "identity/*"
+    {
       capabilities = ["create", "read", "update", "delete", "list", "sudo"]
     }
 
-    # Overwrite `default` policy
-    # ==========================
+    # Create and manage ACL policies broadly across Vault
 
-    # NB: This is necessary because the rule above (path "*") is "weaker" than
-    #     more explicit paths and thus get overriden, typically by the
-    #     `default` policy, which is usually attached as well.
-
-    # NB: The `default` policy can been seen thus:
-    #
-    #     ```sh
-    #     $ vault read sys/policy/default
-    #     ```
-
-    # Allow all actions on the current entity
-    #
-    # NB: The curly brackets contraption is to tell Helm to output a double
-    # curly brackets, which is what Vault also uses for its templated policies.
-    # We also make sure Helm does not insert a newline after, or that will
-    # corrupt the Vault template.
-    path "identity/entity/id/{{ "{{" }}- identity.entity.id }}" {
-      capabilities = ["create", "read", "update", "delete", "list", "sudo"]
+    # List existing policies
+    path "sys/policies/acl"
+    {
+      capabilities = ["list"]
     }
-    path "identity/entity/name/{{ "{{" }}- identity.entity.name }}" {
+
+    # Create and manage ACL policies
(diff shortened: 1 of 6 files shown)


vault/helm/vault/templates/vault-configmap.yaml
```diff
@@ -6,38 +6,71 @@ metadata:
   {{ include "vault-plural.labels" . | nindent 4 }}
 data:
   admin-policy.hcl: |
-    # Allow everything
-    path "*" {
+    # Read system health check
```
fabricetriboix, line 9: @DavidSpek Could you please provide me with a rationale as to why you reverted the admin policy to what it was before my change? I based my change on [this](https://discuss.hashicorp.com/t/vault-admin-policy/39803). Many thanks.

vault/helm/vault/templates/vault-configmap.yaml
```diff
@@ -53,12 +86,18 @@ data:
...
+  write-admin-policy.sh: |
+    echo "Writing admin policy"
+    vault policy write admin /policies/admin-policy.hcl
   {{- if .Values.oidc.enabled }}
   enable-oidc.sh: |
     set -euo pipefail
 
     echo "Checking if OIDC is already enabled"
-    if vault auth list | grep -q ^oidc/; then
+    set +e
```
fabricetriboix, line 97: @DavidSpek I don't understand why you reverted this test to what it was before my change. My change is much more readable and reduces 4 lines of code into one.

vault/helm/vault/templates/vault-configmap.yaml
```diff
@@ -75,12 +114,12 @@ data:
     mount_accessor=$(vault auth list | awk '{ if ($1 == "oidc/") { print $3 } }')
     echo "  mount accessor: $mount_accessor"
 
-    echo "Writing 'admins' group"
-    group_id=$(vault write identity/group name=admins policies=admin type=external | awk '{ if ($1 == "id") { print $2 } }')
-    echo "  group id: $group_id"
+    echo "Writing 'admin' user"
```
fabricetriboix, line 117: @DavidSpek I thought you asked to me to introduce an "admin group", which is what I did. Could you please explain to me why you changed that to a single user? Thanks.

vault/helm/vault/templates/vault-configmap.yaml
DavidSpek: Because it didn't work, this changes actually broke the installation of existing users (which also made all subsequent app deployments fail). When I originally created this script I did something similar to what you were doing, but quickly noticed that it failed in a lot of cases. Which is how I ended up with this script.

vault/helm/vault/templates/vault-configmap.yaml
DavidSpek: A way to expose the admin group configuration to the user was never added, which would mean that unintended users could become admin in vault installations where they shouldn't have had these privileges. Down the road we can create a more configurable way to add users/groups and policies.

vault/helm/vault/templates/vault-configmap.yaml
fabricetriboix: I can see you modified this:
```hcl
    # List, create, update, and delete key/value secrets
    path "secret/*"
    {
      capabilities = ["create", "read", "update", "delete", "list", "sudo"]
    }
``` 
into this:
```
    # List, create, update, and delete key/value secrets
    path "*"
    {
      capabilities = ["create", "read", "update", "delete", "list", "sudo"]
    }
```
which pretty much opens up everything, except for the things with a more specific path you otherwise restricted in this same policy... The comment on the StackOverflow post is about applying least privilege principle, which is not what we want, we wanted to create a "superuser" policy where the user does have the ability to do everything they want.

Anyway, the policy you used and modified is not too bad, I guess. We can always tweak it later if some users complain.

PR #279, opened 2022-06-03
Jitsu

tony0306:
## Changes on the Jitsu exposed ports
Adding port 8001 to have successful liveness and readiness probes 

## Checklist


jitsu/helm/jitsu/templates/deployment.yaml
```diff
@@ -43,6 +43,9 @@ spec:
             - name: config
               containerPort: 8000
               protocol: TCP
+            - name: test
```
michaeljguarino, line 46: is `test` the right name for the port or does it serve a different role within jitsu?

tony0306: I think a better name will be readiness/liveness. I used "test" as this is used by the probes to initially test the url. As this was failing (port was not exposed), k8 was killing the deployment.

michaeljguarino: how about `health`?

tony0306: I changed to health. I was also supposed to change it in here:
          livenessProbe:
            httpGet:
              path: /
              port: health

but on the deployment I'm getting this error. 
Liveness probe errored: strconv.Atoi: parsing "health": invalid syntax

Reverted to http.

jitsu/helm/jitsu/templates/service.yaml
```diff
@@ -15,5 +15,9 @@ spec:
       targetPort: config
       protocol: TCP
       name: config
+    - port: 8001
+      targetPort: 8001
```
DavidSpek, line 19: This should probably be set to `health`

michaeljguarino (approved):
rm the file for n8n but otherwise I think this is good to go

PR #282, opened 2022-06-07
Begin onboarding kubescape

michaeljguarino:
## Summary

This could help seal a partnership with armosec, so lets get it up and ready to go.

## Test Plan
validating in our prod cluster

## Checklist


DavidSpek:
@michaeljguarino I believe they rely on OPA, so we'd need to onboard that as well.

michaeljguarino:
Just tested this in our prod cluster and everything seems kosher

Issue #178, opened 2022-03-04
Onboard Presto


michaeljguarino:
## Use Case
Probably the best horizontally scalable, apache-licensed datawarehouse, esp with superset support.  We'll need internally as well.

## Ideas of Implementation
Need to scour the web for presto-on-k8s implementations

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

jaystary:
Presto has been renamed to Trino btw.

Should consider pairing Trino with Hive, which it uses as a metastore for its SQL Query Engine
https://trino.io/blog/2020/10/20/intro-to-hive-connector.html

We would be interested in using this as well, as we are currently contemplating a data architecture around this although SQL Query ability is not our top prio.

avaidyanatha:
This would be great!

PR #303, opened 2022-06-20
Onboard Trino First version

tony0306:
Closes: https://github.com/pluralsh/plural-artifacts/issues/178
Onboarding Trino

## Summary
Onboarding Trino app. Image should be stored in plural ECR.


DavidSpek:
@tony0306 We should just have a dependency on https://github.com/trinodb/charts rather than copying all the templates into the artifact.

tony0306:
> @tony0306 We should just have a dependency on https://github.com/trinodb/charts rather than copying all the templates into the artifact.

I guess this should be in chart.yaml. I will check to see if I made any changes that will make this an issue.

PR #304, opened 2022-06-23
feat: AWS MinIO distributed directpv

DavidSpek:
## Summary
This PR updates our distributed MinIO deployment on AWS to use [minio directpv](https://github.com/minio/directpv) together with nodes that have local HDDs.

## Test Plan
local linking


minio/helm/directpv/certs/ca.pem
```diff
@@ -0,0 +1,33 @@
+-----BEGIN CERTIFICATE-----
```
michaeljguarino, line 1: should we check this in?

DavidSpek: Sadly I think we have to. Their CLI generates the certificates at installation time, which were for a different namespace so I needed to generate new ones manually. I don't think we'll be able to use cert-manager since the operator is coded to look for `ca.pem`, `cert.pem` and `key.pem` and cert-manager will create `ca.crt`, `tls.crt` and `tls.key`.

If they move away from only being able to deploy directpv with their CLI that would make me very happy haha.

PR #307, opened 2022-06-26
Strapi

gittysachin:
Adding strapi integration https://github.com/pluralsh/plural-artifacts/issues/132. Initial push for testing


strapi/helm/strapi/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
michaeljguarino, line 1: Usually it's a good idea to drop these helm tests as they can make deployments more flaky

strapi/helm/strapi/values.yaml
```diff
@@ -0,0 +1,95 @@
+# Default values for strapi.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: gittysachin/strapi
```
michaeljguarino, line 8: you should be able to reference this as `dkr.plural.sh/strapi/gittysachin/strapi` once the vendor images action runs.

michaeljguarino (commented):
This looks pretty good, some minor suggestions but overall this looks good to ship!

michaeljguarino:
No need to update docs.plural.sh for this btw, we can handle that ourselves (there's a script we've written that does it)

strapi/helm/strapi/values.yaml
gittysachin: @michaeljguarino got it, so should I change the repository name to include `dkr.plural.sh/strapi` now?

strapi/helm/strapi/templates/tests/test-connection.yaml
gittysachin: Removing

michaeljguarino (approved):
This looks pretty good, some minor suggestions but overall this looks good to ship!

strapi/helm/strapi/values.yaml
gittysachin: Changed the image url to `dkr.plural.sh`

PR #313, opened 2022-07-01
Redash

tony0306:
## Summary
Onboarding Redash. PostgreSQL and Redis are included by the redash community chart. 


michaeljguarino:
@tony0306 I think we should actually focus on lightdash not redash

michaeljguarino:
We can still put this in though, there are people who do want it

tony0306:
I think that I'm pretty close to finish it, once done I will move to lightdash as well. I guess next week this might be done as well.

redash/helm/redash/values.yaml.tpl
```diff
@@ -0,0 +1,33 @@
...
+  application:
+    links:
+    - description: redash web ui
+      url: {{ .Values.hostname }}
+
+redash:
+  redash:
+    secretKey: {{ dedupe . "redash.redash.redash.secretKey" (randAlphaNum 32) }}
+    cookieSecret: {{ dedupe . "redash.redash.redash.cookieSecret" (randAlphaNum 32) }}
+  postgresql:
```
michaeljguarino, line 12: these values should be in the base values.yaml not here

redash/helm/redash/values.yaml.tpl
```diff
@@ -0,0 +1,33 @@
...
+    enabled: false
+  redis:
+    enabled: false
+  externalRedisSecret:
```
michaeljguarino, line 16: similarly can go in plain values.yaml

redash/plural/notes.tpl
```diff
@@ -0,0 +1 @@
+Use `plural watch redash` to track the status of your application
```
michaeljguarino, line 1: we should add a line like:

`View your redash instance at {{ .Values.hostname }}`

michaeljguarino (approved):

PR #314, opened 2022-07-04
Add postgres wait init containers to airbyte app

michaeljguarino:
## Summary

We're seeing occasional issues waiting for airbyte migrations which might be due to race conditions with postgres provisioning, this should serialize the process a bit and make it more reliable/debuggable

## Test Plan
tested using plural link

## Checklist


DavidSpek:
Don’t forget to jump the chart version.

michaeljguarino:
ugh yea

PR #327, opened 2022-07-26
Onboard YugabyteDB

DavidSpek:
## Summary
This PR onboard YugabyteDB. A forked helm chart is used so that we can automate the setup of a [multi availability zone](https://docs.yugabyte.com/preview/deploy/kubernetes/multi-zone/eks/helm-chart/) deployment using the region and AZ labels on the Kubernets nodes.

## Test Plan
Local linking

## Checklist

    - [x]  ideally also have at least cpu/mem utilization dashboards for webserver tier of the app
    - [x]  you can use `plural from-grafana` to convert a grafana dashboard found via google to our CRD
    - [x]  inputing secrets
    - [x]  configuring autoscaling
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


DavidSpek:
The setup job to set the admin passwords is exiting with an error even though the password is changed correctly. This still needs to be solved along with adding a runbook for scaling.

PR #331, opened 2022-07-27
add external db support for metabase

mertkayhan:
## Summary

Hello @michaeljguarino @DavidSpek, 

I have made a minor modification to the metabase chart so that we can use cloud sql instead of operator managed postgres. Could you please review this change?

## Test Plan

Tested locally with plural link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/metabase/helm/metabase/Chart.yaml
+++ b/metabase/helm/metabase/Chart.yaml
@@ -8,3 +8,4 @@ dependencies:
 - name: postgres
   version: 0.1.4
   repository: https://pluralsh.github.io/module-library
+  condition: database.operatorEnabled
--- a/metabase/helm/metabase/values.yaml.tpl
+++ b/metabase/helm/metabase/values.yaml.tpl
@@ -4,8 +4,15 @@ global:
     - description: metabase web ui
       url: {{ .Values.hostname }}
 
+{{ if not .Values.postgresqlDisabled }}
 database:
   encryptionKey: {{ dedupe . "metabase.database.encryptionKey" (randAlphaNum 32) }}
+  operatorEnabled: true
+{{ else }}
+database: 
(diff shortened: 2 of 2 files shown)


metabase/helm/metabase/Chart.yaml
```diff
@@ -7,4 +7,5 @@ appVersion: "1.16.0"
 dependencies:
 - name: postgres
   version: 0.1.4
-  repository: https://pluralsh.github.io/module-library
\ No newline at end of file
+  repository: https://pluralsh.github.io/module-library
+  condition: database.operatorEnabled
```
michaeljguarino, line 11: I believe this needs to be set to default on?

michaeljguarino (changes requested):

metabase/helm/metabase/Chart.yaml
mertkayhan: I've set it to true [here](https://github.com/pluralsh/plural-artifacts/pull/331/files#diff-69d83cf79d91da3fc1932425bf0455269314f1b91073c95e41f56a33e2c7007fR10). I figured this would work, what do you think?

michaeljguarino: ah I see yeah that's probably fine

michaeljguarino: let me test on our instance and give the heads up if it's good to merge

michaeljguarino (approved):
Just tested and this looks good to ship!

michaeljguarino:
@mertkayhan bump the version for the chart and I'll merge it

mertkayhan:
> @mertkayhan bump the version for the chart and I'll merge it

@michaeljguarino Thanks for checking it! I did not manage to update the tag yet. Did you do it?

PR #332, opened 2022-07-28
Move MLFlow deployment away from operator

DavidSpek:
## Summary
Sets the annotation needed for MLFlow to visualize artifacts. Still need to build a new image with boto3 and an updated version of MLFlow.


mlflow/plural/recipes/mlflow-aws.yaml
```diff
@@ -25,9 +25,5 @@ sections:
   items:
   - type: TERRAFORM
     name: aws
-  - type: TERRAFORM
-    name: aws-standalone
   - type: HELM
     name: mlflow
```
michaeljguarino, line 29: This means existing users with an mlflow install are going to have the operator uninstalled and this chart replacing it?

michaeljguarino (commented):
We should take a look into the total number of people currently using mlflow.  A breaking cha

mlflow/plural/recipes/mlflow-aws.yaml
DavidSpek: That is the idea. MLFlow standalone will be replaced by this, the operator is no longer needed either.

mlflow/terraform/aws/main.tf
```diff
@@ -4,6 +4,55 @@ resource "kubernetes_namespace" "mlflow" {
...
+  source = "github.com/pluralsh/module-library//terraform/irsa-policy"
+  cluster_name   = var.cluster_name
+  role_name      = var.role_name
+  namespace      = var.namespace
+  serviceaccount = var.serviceaccount_name
+  policy_json    = data.aws_iam_policy_document.mlflow.json
+
+}
+
+data "aws_iam_policy_document" "mlflow" {
```
michaeljguarino, line 28: the s3 buckets module should autocreate a policy for you

mlflow/terraform/gcp/main.tf
```diff
@@ -4,6 +4,7 @@ resource "kubernetes_namespace" "mlflow" {
     labels = {
       "app.kubernetes.io/managed-by" = "plural"
       "app.plural.sh/name" = "mlflow"
+      "platform.plural.sh/sync-target" = "pg"
     }
   }
```
michaeljguarino, line 9: looks like gcp/azure aren't having their object stores provisioned?

DavidSpek: We also don't have recipes for GCP and Azure yet. Given the past experience with MLFlow now we should test those cloud providers rigorously before releasing those recipes.

PR #341, opened 2022-08-06
Add default read-only role to argo-workflow sso

michaeljguarino:
## Summary

Argo has the capability to provide a default role to sso users, which can naturally be made a RO role so users can use the workflow dashboard for visibility.  We should configure that properly by default.

## Test Plan
verified with plural link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/argo-workflows/helm/argo-workflows/Chart.yaml
+++ b/argo-workflows/helm/argo-workflows/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argo-workflows
 description: A Helm chart for Kubernetes
 type: application
-version: 0.1.22
+version: 0.1.23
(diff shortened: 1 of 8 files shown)


DavidSpek:
We should change the name of the SA with the admin role to `argo-workflow-admin` and not add the IRSA and workload identity annotations to it since that SA probably shouldn't run the actual workflows.

For workflows we can create the default `argo-workflow` SA and then use our operator to sync it into other namespaces. The IRSA policy has a wild card for namespaces so copying it over with those annotations should work fine.

michaeljguarino:
I don't think the operator syncs service accounts currently so we'd need to make that change first

PR #348, opened 2022-08-16
bump gitlab to 5.9.0

michaeljguarino:
## Summary
We need to start rolling up gitlab to closer to its current vsn.  This will get us almost to the 6.0.x releases

## Test Plan
tested locally via plural link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


tony0306 (commented):
No visible issue.

PR #356, opened 2022-08-29
adjusted variables accordingly to fix "culling"

tikr7:
Hi @DavidSpek 

We analyzed the issue of the "culling" feature of the notebook controller. In the end the names of the variables have changed (https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220121-jupyter-notebook-idleness.md#api-changes). 

This PR is to fix it.

I already tried it on our cluster by changing the configmap manually and we can confirm that it works.


michaeljguarino (approved):

michaeljguarino (commented):
One thing: you'll need to bump the version on the containing chart to have this publish

tikr7:
Is it now the correct version I need to bump?

@DavidSpek I also adjusted the notebook-controller version to 1.5.0 as agreed in the daily.
Interestingly, it was version 1.3 before as the image.

DavidSpek:
You forgot to bump the version of the webapp image it seems.

DavidSpek:
the image tag for the web app is v1.5.0

tikr7:
Thanks! Updated.

DavidSpek:
Thanks for the contribution!

PR #357, opened 2022-08-31
Onboard terraria

floreks:
## Summary
Closes ENG-518.

Onboarded terraria. Exposed some basic server settings for the user during bundle install:
- `hostname` - fqdn of the server
- `worldsize` - one of: `small|medium|large`
- `password` - protected the server with password
- `restAPIEnabled` - when TShock is used (default) it can also expose the rest API on port 7878.

## Test Plan
Used the `make upload-terraria` and then deployed it on my AWS cluster using plural CLI. Test server available at: `terraria.floreks-aws.onplural.sh:7777`.

Rest API: [status](http://terraria.floreks-aws.onplural.sh:7878/status)

## Checklist
    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


terraria/helm/terraria/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
michaeljguarino, line 1: a lot of these files are autogen'ed from helm create and can just be rm'ed

terraria/helm/terraria/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
michaeljguarino, line 1: probably want to rm this file

terraria/vendor_images
```diff
@@ -0,0 +1 @@
+ryshe/terraria:latest
```
DavidSpek, line 1: Probably shouldn't use latest here since then we can never push out an update to users.

terraria/helm/terraria/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: terraria
+description: helm chart for terraria
+type: application
+version: 0.1.2
```
floreks, line 5: Do i need to bump this version if I have already published it?

DavidSpek: The version needs to not have been pushed to Plural yet.

DavidSpek (changes requested):
I thought the plan was to add an ingress to expose the server, rather than a dedicated loadbalancer.

floreks:
Ingresses and nginx controller are designed to handle the web traffic (http/https). In case of game servers where it requires a raw TCP connection on a specific port it's a bit different and requires too many changes in the nginx controller configuration to be actually worth the effort.

DavidSpek (approved):

DavidSpek:
We could actually do it, since we're doing it for GitLab as well. But for now this should be fine. Although I'm not sure if this will work for our KinD setup. It's a bit ironic that we need load balancers for the game servers, but using load balancers is less easy where the game servers are likely to run (like KinD or on-prem).

floreks:
It might make sense for Gitlab. I am not sure if it does for most game servers though. Many users don't need or want to configure the hostname for their game server and use plain IP to connect to the server or share it with people. At least that's my experience.

PR #358, opened 2022-08-31
AWS: ensure node groups have a unique name

DavidSpek:
## Summary
When an AWS region only has 2 availability zones a duplicate name for our node groups occurs. This PR uses a new version of our node group module that appends the subnet name to the node group name to ensure they are unique. Since this is a breaking change, this is also a good opportunity to update the AMI used by the worker nodes. Also, the `kube-proxy` and `vpc-cni` versions are updated.

## Test Plan
local linking


michaeljguarino (approved):
I think there's also a way to generate notes for this as well. Also how has it been tested so far?

PR #367, opened 2022-09-13
Bug/fix kfp token injection

tikr7:
Hi @DavidSpek 
In order to fix the KFP token injection the following label needs to be set in the user namespaces:
```
  labels:
    app.kubernetes.io/part-of: kubeflow-profile
```
In a commit in ~May from you, that label got deleted.

Here in this PR I added it again. And I bumped the version.


kubeflow/helm/profile-controller/values.yaml
```diff
@@ -70,6 +70,7 @@ config:
       bucketName: ""
```
DavidSpek, line 70: Can you bump the image tag to 0.1.13.

PR #375, opened 2022-09-20
Use forked postgres operator with liveness probes and fix backup lifecycle

DavidSpek:
## Summary
This PR updates the postgres operator to the latest version along with a modification that allows for setting liveness probes for databases. This should fix issues where patroni can't talk to the Kubernetes API anymore. It also uses a newer Spilo image that fixes the backup retention policy which wasn't working. Another small fix is that the operator will now add the necessary annotation for IRSA to service accounts it creates for the postgres pods (so this doesn't need to be done with terraform anymore).

## Test Plan
Local linking. 
- Check that backups are pushed to S3.
- Check that WAL archives are pushed to S3.
- Check that old backups are cleaned up.
- Check that there aren't any errors in the `pg_log` logs (related to send WAL archives to S3)

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/postgres/helm/postgres/values.yaml
+++ b/postgres/helm/postgres/values.yaml
@@ -10,7 +10,7 @@ oauthSecretName: postgres-operator
 image:
   registry: dkr.plural.sh
   repository: postgres/acid/postgres-operator
-  tag: v1.8.2-6-g8c94e377-dirty
+  tag: v1.8.2-7-g73a0dfee-dirty
   pullPolicy: "IfNotPresent"
 
   # Optionally specify an array of imagePullSecrets.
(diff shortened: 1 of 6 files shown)


postgres/helm/postgres/values.yaml
```diff
@@ -8,9 +8,9 @@ operatorConfigName: operator-configuration
 oauthSecretName: postgres-operator
 
 image:
-  registry: registry.opensource.zalan.do
-  repository: acid/postgres-operator
-  tag: v1.6.3
+  registry: dkr.plural.sh
```
michaeljguarino, line 11: could we use a gcr or ghcr repo for this?

DavidSpek: We could, but I thought we wanted to unify using the plural registry so people know where all images come from and that they have. been scanned for vulnerabilities. This would also be needed for vulnerability based upgrade policies.

PR #383, opened 2022-10-12
Ray: add support for git-sync

DavidSpek:
## Summary
Allow for easily adding a git-sync sidecar container for the ray head server to sync in python code that should be executed.

## Test Plan
Local linking.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/ray/helm/ray/values.yaml.tpl
+++ b/ray/helm/ray/values.yaml.tpl
@@ -50,3 +50,16 @@ oauth2-proxy:
 users:
 {{ toYaml .Values.users | nindent 2 }}
 {{ end }}
+
+{{- if .Values.gitSync.enabled }}
+gitSync:
+  enabled: true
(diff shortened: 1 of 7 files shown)


ray/helm/ray/values.yaml.tpl
```diff
@@ -50,3 +50,16 @@ oauth2-proxy:
...
 {{ toYaml .Values.users | nindent 2 }}
 {{ end }}
+
+{{- if .Values.gitSync.enabled }}
+gitSync:
+  enabled: true
+  repo: {{ .Values.gitSync.repo }}
+  {{- if .Values.gitSync.password }}
+  password: {{ .Values.gitSync.password }}
+  username: {{ .Values.gitSync.username }}
```
michaeljguarino, line 60: we should default this to the `apiKey` name github users for PATs

PR #397, opened 2022-10-21
airbyte use postgres module chart + cleanup

DavidSpek:
## Summary
This PR migrates our Airbyte installation to use our Postgres module chart rather than templating the postgres CRD in this chart directly, allowing for easy upgrades and maintenance in the future. It also cleans up the templating for people that want to disable postgres.

## Test Plan
Local linking.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


airbyte/helm/airbyte/values.yaml.tpl
```diff
@@ -41,20 +41,12 @@ private:
 
 {{ $minioNamespace := namespace "minio" }}
 
-{{ if not .Values.postgresDisabled }}
-airbyte:
-  postgresql:
```
michaeljguarino, line 46: we should probably keep this to be safe.  I've had weird behavior with helm values before

DavidSpek: It's on line 44-46 and 49 now. Needed to move it to another location since the value is in a different spot. Since the location has changed existing values in the `values.yaml` in installation repos won't get touched. I've tested the upgrade and nothing strange happened to the values file.

PR #403, opened 2022-10-24
select subnet route table for s3 vpc endpoint

ncbkr:
## Summary

earlier today the PR #400 was merged. The purpose was to route traffic from within the cluster to S3 without leaving the aws network. I wanted to make sure the PR really rsolves the issue and looked for a way to validate that the route to S3 does not involve any nodes on the public internet.

I found [this](https://aws.amazon.com/premiumsupport/knowledge-center/vpc-fix-gateway-or-interface-endpoint/?nc1=h_ls) article on how to verify the setup.

Basically, it says that the output of tcptraceroute should not contain any hops. So it should consist only of line with `*` characters like e.g. line 3 in the below screen shot. This screenshot also shows several lines which contain IP addresses. Not sure what those ip addresses belong to, but according to the above article, those lines should not be there.

image

In the PR #400 we provided the id of the main route table of the VPC containing our EKS cluster to the terraform configuration of the s3 endpoint.

This new PR uses the id of the route table for the `<cluster_name>-worker-private-<region>` subnets instead. These subnets are also used when creating the `single_az_node_groups` in the `bootstrap` artifact.

After applying the change, I ran tcptraceroute one more time and got the result shown in the screenshot below. This corresponds to the expected result shown in the above article.

image

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/bootstrap/terraform/aws-bootstrap/s3-vpc-endpoint.tf
+++ b/bootstrap/terraform/aws-bootstrap/s3-vpc-endpoint.tf
@@ -1,6 +1,10 @@
+data "aws_route_table" "worker_private_subnets_route_table" {
+  subnet_id = module.vpc.worker_private_subnets_ids[0]
+}
+
 resource "aws_vpc_endpoint" "s3" {
   vpc_id       = module.vpc.vpc_id
   service_name = "com.amazonaws.${var.aws_region}.s3"
   auto_accept = true
-  route_table_ids = [module.vpc.default_route_table_id]
+  route_table_ids = [data.aws_route_table.worker_private_subnets_route_table.id]
 }


michaeljguarino:
Have you tested this locally?

ncbkr:
I deployed this change to our cluster using our up-to-date fork of the plural-artifacts repository and `plural link` for the aws-bootstrap terraform module. I tested this for the `eu-central-1` region as described above.

Happy to do more extensive testing if required.

DavidSpek (approved):
Thanks for the contribution, looks good!

PR #421, opened 2022-11-02
External Secrets Operator Onboarding

BogdanAntoniu78:
Onboarding external secrets operator

## Checklist

- [X]  No images hosted from dockerhub
    - [ ]  inputing secrets
    - [ ]  Azure
    - [X]  AWS
    - [ ]  GCP


michaeljguarino (approved):

BogdanAntoniu78:
Prerequisites:

- Policy to access AWS Secret Manager (included in terraform main.tf)
- Role to be assumed via IRSA, with trust relationship with EKS OIDC provider and permission based on the policy created above
- output.tf should contain account_id and the service account name included in OIDC qualified subjects
- k8 values.yaml.tpl should include ServiceAccount variable with the correct annotation including the account id and the role to be assumed (defined above)

PR #425, opened 2022-11-04
Backfill GCP and Azure support for dagster

michaeljguarino:
## Summary

It was brought to my attention we never rounded out gcp and azure for dagster, which is a tad embarrassing.  This should handle it, the main issue is getting object storage working.

## Test Plan
local

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


BogdanAntoniu78:
There might some other apps as well actually.

DavidSpek:
Hold as I'm just about to get the dagster upgrade merged in. Then this will need rebasing.

dagster/terraform/azure/main.tf
```diff
@@ -8,3 +8,26 @@ resource "kubernetes_namespace" "dagster" {
     }
   }
 }
+
```
DavidSpek, line 11: Can you rename this TF module from `azure` to `minio`, this way it can also be used by other providers in the future without it being confusing or duplicating the code.

michaeljguarino: I think that might rip out some existing installations unfortunately

PR #427, opened 2022-11-06
Adding docker-in-docker container into jitsu deployment

Alessar:
Summary:
Adding docker-in-docker-daemon.
It helps **jitsu** to get access to sources through **airbyte**

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/jitsu/helm/jitsu/templates/deployment.yaml
+++ b/jitsu/helm/jitsu/templates/deployment.yaml
@@ -71,7 +71,7 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
         {{- if .Values.airbyte.enabled }}
         - name: dind-daemon
-          image: docker:dind
+          image: {{ .Values.airbyte.dindImage }}
           command: [ "dockerd", "--host", "tcp://127.0.0.1:2375" ]
           securityContext:
             privileged: true
--- a/jitsu/helm/jitsu/templates/deployment.yaml
+++ b/jitsu/helm/jitsu/templates/deployment.yaml
@@ -28,11 +28,13 @@ spec:
       serviceAccountName: {{ include "jitsu.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- if .Values.airbyte.enabled }}
       volumes:
         - name: docker-graph-storage
           emptyDir: {}
         - name: jitsu-workspace
           emptyDir: {}
+      {{- end }}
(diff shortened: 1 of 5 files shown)


jitsu/helm/jitsu/templates/deployment.yaml
```diff
@@ -57,6 +57,11 @@ spec:
             periodSeconds: 5
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+        - name: dind-daemon
```
michaeljguarino, line 60: I think to run docker in docker on a cloud k8s deployment, you need more since the host docker socket isn't exposed.   I've done this before and think you need something like:

```
 securityContext:
    privileged: true
  env:
  - name: DOCKER_TLS_CERTDIR
    value: /certs
  volumeMounts:
  - name: share
    mountPath: /tmp/shared
  - name: docker-storage
    mountPath: /var/lib/docker
  - name: docker-certs
    mountPath: /certs
volumes:
- name: share
  emptyDir: {}
- name: docker-storage
  emptyDir: {}
- name: docker-certs
  emptyDir: {}
```

And then the other container will need to set a DOCKER_HOST env var to tcp://localhost:2375

michaeljguarino: we'd probably want this to be optional too, since some people won't want the security issues with privileged containers/dind generally

Alessar: @michaeljguarino Thanks for providing the configuration part.
Some parts of the configuration were updated.
Testing on AWS showed that it is working.

michaeljguarino: the certs dir didn't need to be mounted?

Alessar: It is an interesting question. I tested only our purposes.
For our purposes, I did not find issues with certificates without that directory.
But maybe there is some behaviour that is out of our scope.

michaeljguarino (approved):
Bump the version in `Chart.yaml` so it can be applied, but it looks good to me.  Thinking about it some more, there's really no major downside of having this as the default, and if someone complains, we can make it optional then.

jitsu/helm/jitsu/templates/deployment.yaml
Alessar: New options for the new capability were added.

PR #433, opened 2022-11-10
Upgrade Airbyte to v0.40.18

pettersamuelsen:
## Summary
Upgrade AirByte to v.0.40.18

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino:
We did this same work in a separate PR that handled some architectural differences in the versions, so should no longer be necessary

PR #437, opened 2022-11-13
Vault private ingress

BogdanAntoniu78:
## Summary
Changes on this PR:
- Private ingress  added for Vault deployment
- Runbooks 
- CPU/Memory requests and limits set 

## Test Plan
- Runbooks corectly displayed in Console
- Pods are starting with new requests and limits
- If exposePrivate flag is set to true, Vault UI is available only when using VPN connection to EKS cluster 

## Checklist

- [X]  No images hosted from dockerhub
    - [X]  ideally also have at least cpu/mem utilization dashboards for webserver tier of the app
- [X]  Are scaling runbooks present
    - [ ]  inputing secrets
    - [X]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/vault/helm/vault/values.yaml.tpl
+++ b/vault/helm/vault/values.yaml.tpl
@@ -43,7 +43,11 @@ vault:
         cert-manager.io/cluster-issuer: letsencrypt-prod
         nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
         nginx.ingress.kubernetes.io/use-regex: "true"
+      {{- if not .Values.exposePrivate }}
       ingressClassName: nginx
+      {{- else }}
+      ingressClassName: internal-nginx
+      {{- end }}
       hosts:
       - host: {{ .Values.hostname }}
         paths: []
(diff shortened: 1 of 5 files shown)


vault/helm/vault/values.yaml.tpl
```diff
@@ -43,7 +43,11 @@ vault:
         cert-manager.io/cluster-issuer: letsencrypt-prod
         nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
         nginx.ingress.kubernetes.io/use-regex: "true"
+      {{- if not .Values.exposePrivate }}
       ingressClassName: nginx
```
michaeljguarino, line 47: A lot of these values should actually be moved down to values.yaml since the current setup prevents reconfiguration.  Really we should have nginx as the default there then overwrite with:

{{ if .Values.exposePrivate }}
ingressClassName: internal-nginx
{{ end }}

in `values.yaml.tpl`

BogdanAntoniu78: Sure. I moved the values to values.yaml.tpl. Ihad to leave also the else statement, as otherwise the className gets undefined.

michaeljguarino (approved):

michaeljguarino:
It looks like this does have a merge conflict to resolve

PR #443, opened 2022-11-16
feat: update airbyte and use upstream chart

DavidSpek:
## Summary
This PR migrates our Airbyte setup to use the upstream helm chart, along with a change to allow for adding labels to pods. Once https://github.com/airbytehq/airbyte/pull/19474 is merged we should be able to use the upstream airbyte helm chart directly.

A lot has changed in the chart structure, and with it the way Airbyte is configured. Using the upstream chart will hopefully make is so that we don't run into errors due to misconfiguration of Airbyte because our forked chart isn't up to date.

## Test Plan
Local linking.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


DavidSpek:
This is blocked by us needing to push out a dummy airbyte update once the new console is released so we can migrate `values.yaml` into `default-values.yaml` before this upgrade is done.

PR #445, opened 2022-11-21
Upgrade Azure AKS too 1.23.12 + configure autoscaler settings

davidspek:
## Summary
This PR upgrades our AKS deployment to Kubernetes 1.23.12. It also disables the ingress controller, external dns and OPA that was being deployed and managed by Azure. Furthermore, it also updates our terraform module to be able to configure settings for the cluster autoscaler managed by Azure.

## Test Plan
Local linking

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino:
How much testing have we done for this so far?

DavidSpek:
It's what's deployed on our Azure cluster currently. The main thing that we need to be careful with is not destroying Azure clusters with invalid changes, which isn't the case here.

PR #476, opened 2022-12-29
Support BYOK for aws EKS

michaeljguarino:
## Summary

The approach here is actually kind of simple, just allow the aws-bootstrap module to reconfigure for existing eks installs and rewire a few outputs. All other modules assume it's created by aws-bootstrap so they just work for free. There are still some complications:

* cluster autoscaling might not work if nodes are misconfigured
* can't know for sure whether the cluster has an oidc provider for IRSA, so need a variable to disable it.

## Test Plan
* created a cluster with eksctl, ran terraform init, modified tf and deployed airbyte, confirmed it worked and could also be destroyed
* linked against prod repo, confirmed terraform plan showed no changes

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


bootstrap/terraform/aws-bootstrap/aws-lb-controller.tf
```diff
@@ -3,14 +3,14 @@ module "assumable_role_alb" {
...
   role_name                     = "${var.cluster_name}-alb"
-  provider_url                  = replace(module.cluster.cluster_oidc_issuer_url, "https://", "")
+  provider_url                  = replace(local.cluster_oidc_issuer_url, "https://", "")
   role_policy_arns              = [aws_iam_policy.alb.arn]
   oidc_fully_qualified_subjects = ["system:serviceaccount:${var.namespace}:${var.alb_serviceaccount}"]
 }
 
 resource "aws_iam_policy" "alb" {
   name_prefix = "alb-contrller"
-  description = "aws load balancer controller policy for cluster ${module.cluster.cluster_id}"
```
DavidSpek, line 13: There's a pretty high likelihood the LB controller is already deployed on clusters so we should make this optional.

bootstrap/terraform/aws-bootstrap/existing.tf
```diff
@@ -0,0 +1,9 @@
+data "aws_eks_cluster" "cluster" {
+    count = var.create_cluster ? 0 : 1
+    name = var.cluster_name
```
DavidSpek, line 3: Our cluster name variable is limited to 15 characters through the CLI, but they can be longer in AWS. When deploying a cluster through the Cluster-API it will add some prefixes to cluster names so they will be longer than our CLI supports. We should allow users to select from existing clusters if they choose to use BYOK in our CLI during the `plural init` stage.

bootstrap/terraform/aws-bootstrap/locals.tf
```diff
@@ -0,0 +1,13 @@
+locals {
+    sts_principal = "sts.${data.aws_partition.current.dns_suffix}"
+
+    private_subnet_ids = var.create_cluster ? module.vpc.private_subnets_ids : var.private_subnet_ids
+    public_subnet_ids = var.create_cluster ? module.vpc.public_subnets_ids : var.public_subnet_ids
```
DavidSpek, line 5: Do we need these variables if we can get them from the cluster data object already?

DavidSpek (commented):
Some initial comments of things that can be improved.

PR #485, opened 2023-01-10
add lua support

zreigz:
## Summary
Migrate grafana go template to Lua script

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


grafana/helm/grafana/values.yaml.lua
```diff
@@ -0,0 +1,119 @@
+output = {
+    global={
+        application={
```
michaeljguarino, line 3: we should definitely have helper functions for these

michaeljguarino (approved):
this is definitely better, but a part of me wonders if javascript is a bit better from a dev ux perspective.  Let's see how this goes and potentially work on embedding js

PR #488, opened 2023-01-16
remove job needed for airbyte upgrade

DavidSpek:
## Summary
The airbyte upgrade needed a job to remove the deployments since immutable fields were being changed. This PR removes that job so that future chart upgrades won't remove all deployments and then redeploy them.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino (approved):

airbyte/helm/airbyte/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: airbyte
 description: Unified data integration platform
 type: application
-version: 0.4.0
+version: 0.4.1
```
michaeljguarino, line 5: this will need to be bumped again

PR #498, opened 2023-01-26
feat(postgres): enable AES256 bucket encryption on AWS

DavidSpek:
## Summary
This PR enables AES256 encryption of the S3 bucket used for base backups and wal archives of postgres clusters managed by the operator.

## Test Plan
Deployed using local linking and checked that current base backups and wal archives are still appearing in the S3 bucket as expected.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/postgres/terraform/aws/variables.tf
+++ b/postgres/terraform/aws/variables.tf
@@ -21,3 +21,9 @@ variable "role_name" {
   type = string
   default = "postgres"
 }
+
+variable "force_destroy_bucket" {
+  type        = bool
+  default     = false
+  description = "If true, the bucket will be deleted even if it contains objects."
+}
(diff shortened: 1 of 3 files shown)


postgres/terraform/aws/variables.tf
```diff
@@ -21,3 +21,9 @@ variable "role_name" {
   type = string
   default = "postgres"
 }
+
+variable "force_destroy_bucket" {
+  type        = bool
+  default     = false
```
michaeljguarino, line 27: this should still default to true, but we should override it in our cluster's installation

DavidSpek: done

michaeljguarino (approved):

PR #504, opened 2023-01-29
Onboard standalone prefect agent

michaeljguarino:
## Summary

The folks at prefect requested this as they also want people to have a quickstart to use prefect cloud as well. Getting this started off here.

## Test Plan
will test post-merge

## Checklist

    - [x]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


prefect-agent/helm/prefect-agent/deps.yaml
```diff
@@ -0,0 +1,26 @@
...
+  application: true
+  description: Deploys prefect-agent crafted for the target cloud
+spec:
+  dependencies:
+  - type: helm
+    name: bootstrap
+    repo: bootstrap
+    version: '>= 0.5.1'
+  - type: terraform
+    name: aws
```
DavidSpek, line 13: this should be renamed to `kube`.

prefect-agent/helm/prefect-agent/deps.yaml
```diff
@@ -0,0 +1,26 @@
...
+    repo: prefect-agent
+    version: '>= 0.1.0'
+    optional: true
```
DavidSpek, line 16: It shouldn't be optional.

prefect-agent/helm/prefect-agent/deps.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - type: terraform
+    name: azure
+    repo: prefect-agent
+    version: '>= 0.1.0'
+    optional: true
+  - type: terraform
+    name: gcp
+    repo: prefect-agent
+    version: '>= 0.1.0'
+    optional: true
```
DavidSpek, line 26: These should be removed.

prefect/helm/prefect/Chart.yaml
```diff
@@ -19,3 +19,7 @@ dependencies:
   version: 0.1.6
   repository: https://pluralsh.github.io/module-library
   condition: oidc-config.enabled
+- name: test-base
+  repository: https://pluralsh.github.io/module-library
+  version: 0.1.3
```
DavidSpek, line 24: You'll want to use `0.1.5` as that's the latest version.

prefect/helm/prefect/values.yaml
```diff
@@ -27,6 +27,12 @@ prefect-orion:
       postgresql: 5432
     externalHostname: plural-postgres-prefect
 
+test-base:
+  enabled: false
+  repository: prefect
+  testName: prefect-integration
+  promoteTag: warm
+
```
DavidSpek, line 35: You'll need to specify the steps with the latest version of the test-base helm chart. See https://github.com/pluralsh/module-library/blob/main/helm/test-base/values.yaml.

prefect-agent/helm/prefect-agent/Chart.yaml
```diff
@@ -0,0 +1,15 @@
...
+version: 0.1.0
+appVersion: v2.7.9
+dependencies:
+- name: prefect-agent
+  version: 2023.1.20
+  repository: https://prefecthq.github.io/prefect-helm
+  condition: prefect-agent.enabled
+- name: test-base
+  repository: https://pluralsh.github.io/module-library
+  version: 0.1.3
```
DavidSpek, line 14: You'll want to use `0.1.5` here.

prefect-agent/helm/prefect-agent/values.yaml
```diff
@@ -0,0 +1,26 @@
...
+        cpu: 10m
+        memory: 128Mi
+      limits:
+        memory: 256Mi
+
+test-base:
+  enabled: false
+  repository: prefect-agent
+  testName: prefect-agent-integration
+  promoteTag: warm
```
DavidSpek, line 26: You'll need to specify steps here with the latest version of the `test-base` chart. See https://github.com/pluralsh/module-library/blob/main/helm/test-base/values.yaml.

DavidSpek (changes requested):
Few small comments. Overall, would we want this to be a separate application or just a separate recipe in the existing `prefect` application targeting the `prefect-agent` helm chart?

michaeljguarino:
Prefect explicitly wanted a separate listing, that's why we're doing it

prefect/helm/prefect/values.yaml
michaeljguarino: we should modify the latest version so these are the default.  I avoided using it because i didn't want to have to specify them

michaeljguarino: and more importantly, there's a chance prefect will own the packaging here, so a smoother experience is better.

DavidSpek: We can try and streamline the upstream test chart a bit more. But they’ll need to have the newer chart to be able to run useful tests.

PR #523, opened 2023-02-10
Add docs for configuring trino catalogs

michaeljguarino:
## Summary
This is a manual step so if we can at least point to it, that could help

## Test Plan
n/a

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/airflow/helm/airflow/values.yaml
+++ b/airflow/helm/airflow/values.yaml
@@ -27,8 +27,8 @@ postgres:
 
 exporter:
   image:
-    repository: gcr.io/pluralsh/prom/statsd-exporter
-    tag: v0.20.2
+    repository: dkr.plural.sh/airflow/prom/statsd-exporter
+    tag: v0.22.8
     pullPolicy: IfNotPresent
   
   podSecurityContext: {}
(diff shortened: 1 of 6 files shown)


airflow/helm/airflow/values.yaml
```diff
@@ -27,8 +27,8 @@ postgres:
 
 exporter:
   image:
-    repository: gcr.io/pluralsh/prom/statsd-exporter
-    tag: v0.20.2
```
DavidSpek, line 31: the tag should be `v0.23.0` based on the image vendoring.

airflow/helm/airflow/values.yaml
```diff
@@ -54,6 +54,7 @@ exporter:
     additionalLabels: {}
 
   statsd:
+    relayAddress: "datadog.datadog:8125"
```
DavidSpek, line 57: Shouldn't this be optional and only set if datadog is installed?

DavidSpek (approved):
One small change, other than that it looks good.

airflow/helm/airflow/values.yaml
michaeljguarino: it's not in our registry yet I think

airflow/helm/airflow/values.yaml
michaeljguarino: udp is fire and forget so it should be fine to put it in

PR #526, opened 2023-02-17
Restore istio artifact

floreks:
## Summary

Looks like it was accidentally removed. I will also push this latest version `0.1.100` as it removes dependency on `grafana-tempo`. It should fix other artifacts that were using it.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203886557126845


DavidSpek (approved):
LGTM

PR #532, opened 2023-02-23
remove namespace alert label

DavidSpek:
## Summary
The extra label being added to alerts breaks the ability to use the namespace label to sort alerts coming from a particular namespace. This PR removes the added namespace label from alerts so it is possible to filter alerts coming from a particular namespace.

image

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/monitoring/helm/monitoring/values.yaml
+++ b/monitoring/helm/monitoring/values.yaml
@@ -140,8 +140,6 @@ kube-prometheus-stack:
     enabled: false
 
   defaultRules:
-    additionalRuleLabels:
-      namespace: monitoring
     rules:
       kubeProxy: false
 
--- a/monitoring/helm/monitoring/Chart.yaml
+++ b/monitoring/helm/monitoring/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: monitoring
 description: A Helm chart for Kubernetes
 type: application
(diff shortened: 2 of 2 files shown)


monitoring/helm/monitoring/values.yaml
```diff
@@ -140,8 +140,6 @@ kube-prometheus-stack:
     enabled: false
 
   defaultRules:
-    additionalRuleLabels:
```
michaeljguarino, line 143: I think this might be needed to visualize these alerts in the console currently, I'm not entirely sure though

DavidSpek: We need to improve the way the console gets alerts anyways so that it can properly filter them for namespace, pod name, container name, etc.

michaeljguarino: It already autolinks to an app if the namespace is there, so if removing this just repins them to the original app it might just work

DavidSpek: The namespace label gets added by the alert so this just ensure it doesn't get overridden causing the namespace label to be `monitoring` for every alert.

michaeljguarino (approved):

PR #570, opened 2023-03-14
fix: fix sftpgo ingress path + add optional service to access TCP services

Walkoss:
## Summary
- Fixed sftpgo ui host
- Added optional Service (LoadBalancer) to access TCP services through a domain name


Walkoss:
Do I need to update chart version? Do I increase the patch or minor version as this PR contains a fix and a new feature 😅

PR #572, opened 2023-03-14
Feat/sftpgo wi

Walkoss:
## Summary
- Add sftpgo workload identity
- Fix sftp ui ingress path
- Add optional Service (LoadBalancer) to access TCP services through a domain name 

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [x]  GCP

### DIFF
--- a/sftpgo/terraform/gcp/variables.tf
+++ b/sftpgo/terraform/gcp/variables.tf
@@ -1,0 +1,15 @@
+variable "namespace" {
+  type = string
+  default = "sftpgo"
+}
+
+variable "cluster_name" {
+  type = string
+}
+
+variable "project_id" {
+  type = string
+  description = <<EOF
+The ID of the project in which the resources belong.
+EOF
+}
--- a/sftpgo/helm/sftpgo/deps.yaml
+++ b/sftpgo/helm/sftpgo/deps.yaml
@@ -17,3 +17,9 @@ spec:
     name: kube
     repo: sftpgo
     version: '>= 0.1.0'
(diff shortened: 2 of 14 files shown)


Walkoss:
@DavidSpek my output variable doesn't seem to be replaced in `default-values.yaml`
Do I need to upgrade chart version?
This PR add commits from a previous opened and ready to review [PR](https://github.com/pluralsh/plural-artifacts/pull/570)

sftpgo/terraform/gcp/variables.tf
```diff
@@ -0,0 +1,21 @@
...
+  type = string
+  description = <<EOF
+The ID of the project in which the resources belong.
+EOF
+}
+
+variable "roles" {
+  type        = list(string)
+  description = "A list of roles to be added to the sftpgo workload identity service account"
+  default     = []
```
michaeljguarino, line 20: should you add a default here?

Walkoss: No default value is required initially. Only needed for the GCS output destination. 
Should it be added even if it is not used or let users assign it?

michaeljguarino (approved):
This is probably good to go, just handle the merge conflicts

Walkoss:
@michaeljguarino all good, should we upgrade chart version?

michaeljguarino:
@Walkoss yes definitely upgrade the chart vsn if there are any changes (otherwise it won't be pushed)

PR #574, opened 2023-03-17
Upgrade Datahub to v0.10.0

DavidSpek:
## Summary
This PR updates datahub to the latest version, along with cleaning up some of the jobs we copied locally to be able to set resources.

## Test Plan
Local linking

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino (commented):
Are we confident we don't need those copies anymore?

DavidSpek:
For the kafka and elasticsearch setupjobs they were only there so we could set different resources. The upgrade job we also didn't change anything about so we don't need it in our wrapper chart.

michaeljguarino (approved):

PR #577, opened 2023-03-18
feat: update dex application on Plural

pacoguzman:
## Summary

Bump dex helm chart to latest [version](https://github.com/dexidp/helm-charts) 

## Test Plan

I wasn't able to test it as per the problem indicated [here](https://discord.com/channels/880830238723047424/880837182389108766/1083784743226716231) based on the following error

```sh
plural link helm dagster --path ../plural-artifacts/dagster/helm --name dagster
2023/03/18 11:08:15 open /Users/pacoguzman/dev/github/pacoguzman/pluralsh-repo/dagster/manifest.yaml: no such file or directory
```
Saying that I'm willing to test it throughly when added to the catalog and the manifest.yaml will be avaible. Please let me know if I took wrong assumptions about the issue.

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


michaeljguarino:
Hey Paco, were you interested in using dex?  We left it dormant because we couldn't think of an obvious use case but are happy to change our mind on it

michaeljguarino:
As far as the error you, you just didn't dive all the way to the right dagster chart,  it would need to be `../plural-artifacts/dagster/helm/dagster` (needs to be the dagster chart w/in the helm folder)

PR #589, opened 2023-03-22
fix datahub setup jobs

DavidSpek:
## Summary
Apparently the DNS resolution for the kafka setup job can have issues. This should fix it.

## Test Plan
Fresh deploy using local linking

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/datahub/helm/datahub/values.yaml
+++ b/datahub/helm/datahub/values.yaml
@@ -145,7 +145,7 @@ datahub:
   kafkaSetupJob:
     enabled: true
     image:
-      repository: gcr.io/pluralsh/linkedin/datahub-kafka-setup
+      repository: dkr.plural.sh/datahub/linkedin/datahub-kafka-setup
       tag: v0.10.0
     resources:
       requests:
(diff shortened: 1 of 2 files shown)


michaeljguarino (approved):

datahub/helm/datahub/values.yaml
```diff
@@ -63,8 +72,24 @@ datahub:
     serviceMonitor:
       create: true
     extraInitContainers:
+    - name: wait-for-elasticsearch-job
+      image: gcr.io/pluralsh/bitnami/kubectl:1.25.8
+      imagePullPolicy: IfNotPresent
+      command: [ "/bin/sh", "-c", "until kubectl wait --for=condition=complete --timeout=1s -n datahub job/datahub-elasticsearch-setup-job; do echo 'waiting for datahub-elasticsearch-setup-job'; sleep 1; done" ]
```
michaeljguarino, line 78: we had a container that wrapped this before that we had used I can't remember the name though

michaeljguarino (approved):

PR #592, opened 2023-03-24
Fix growthbook vendoring

michaeljguarino:
## Summary
growthbook doesn't use proper semantic versioning, so we need this hack

## Test Plan
n/a

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


DavidSpek:
We should use their tagged images. See https://hub.docker.com/layers/growthbook/growthbook/1.9.0/images/sha256-e0216d524ccdb63a653c1def66f3810605d3734e819236d2879364e63d5c330d?context=explore.

PR #597, opened 2023-03-27
feat(airbyte-gcp): add airbyte workload identity sa

Walkoss:
## Summary
- Added GCP SA bound to default K8S sa on airbyte namespace to be able to add specific roles

## Test Plan
- Added a BigQuery data editor and a BigQuery user to the roles of this new sa and tested a job sync on airbyte with BigQuery as destination

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [x]  GCP


Walkoss:
One question though, should we use the existing GCP Plural Admin service account instead or create a new one as was done here?

michaeljguarino (approved):

PR #604, opened 2023-03-30
feat: add notes.tpl & ingress to tier

gordalina:
## Summary

This PR fixes the [failed deployment](https://github.com/pluralsh/plural-artifacts/actions/runs/4559430914/jobs/8043383803) because it doesn't have notes.tpl. It also adds ingress to tier.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/tier/plural/recipes/tier-azure.yaml
+++ b/tier/plural/recipes/tier-azure.yaml
@@ -4,10 +4,15 @@ provider: AZURE
 dependencies:
   - repo: bootstrap
     name: azure-k8s
+  - repo: ingress-nginx
+    name: ingress-nginx-azure
 
 sections:
   - name: tier
-    configuration: []
+    configuration:
+      - name: hostname
+        documentation: FQDN to use for your tier installation
+        type: DOMAIN
     items:
       - type: TERRAFORM
         name: kube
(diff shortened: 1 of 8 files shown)


tier/plural/recipes/tier-azure.yaml
```diff
@@ -4,10 +4,15 @@ provider: AZURE
 dependencies:
   - repo: bootstrap
     name: azure-k8s
+  - repo: ingress-nginx
+    name: ingress-nginx-azure
 
 sections:
   - name: tier
-    configuration: []
+    configuration:
```
michaeljguarino, line 12: so you'd probably need to add the helm values configuration for this in `tier/helm/tier/values.yaml.tpl` to wire up the ingress appropriately.  I think `strapi` is a pretty good model of this which just uses the helm starter templates to configure a simple stateless app w/ an ingress on top

michaeljguarino:
Also how does tier persist its data?  Does it need a database, I'd assume it has to be saved somewhere.

gordalina:
> Also how does tier persist its data? Does it need a database, I'd assume it has to be saved somewhere.

All data is stored stripe.

tier/plural/recipes/tier-azure.yaml
gordalina: Pushed a commit to address this.

michaeljguarino (approved):

PR #605, opened 2023-03-30
feat: tier with protected ingress

gordalina:
## Summary

Finished implementation for tier.

A few questions:

1. Is there an example dashboard that I can re-use for tier?
2. I'm not sure I've addressed the secrets correctly, both the stripe API key that needs to be input as well as the basic auth key that's generated upon install

## Test Plan

1. Create a live and test key in stripe
3. Configure tier with live=false and use the live stripe key
4. Configure tier with live=true and use the test stripe key
5. If the pod runs, it means it can talk to stripe correctly. Alternatively you can run `curl -u "{{.Values.tier.basicAuth }}:" https://{{ .Values.hostname }}`

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP

### DIFF
--- a/tier/helm/tier/templates/deployment.yaml
+++ b/tier/helm/tier/templates/deployment.yaml
@@ -33,17 +33,26 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - "tier"
+            {{- if .Values.tier.live }}
+            - "-live"
+            {{- end }}
+            - "serve"
+          envFrom:
+            - secretRef:
+                name: {{ include "tier.fullname" . }}-env
           ports:
             - name: http
-              containerPort: 80
+              containerPort: 8080
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: /
+              path: /v1/whoami
               port: http
           readinessProbe:
             httpGet:
-              path: /
+              path: /v1/whoami
               port: http
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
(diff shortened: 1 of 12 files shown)


tier/helm/tier/templates/deployment.yaml
```diff
@@ -33,17 +33,26 @@ spec:
...
+                name: {{ include "tier.fullname" . }}-env
           ports:
             - name: http
-              containerPort: 80
+              containerPort: 8080
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: /
+              path: /v1/whoami
```
michaeljguarino, line 51: is this a suitable healthcheck?  Not sure if it ends up being an expensive op or not

michaeljguarino (approved):
This should be good to go actually.  The stripe key will be stored securely since the values files are encrypted in git, so no worry there.

As far as dashboards/runbooks, my guess is the only thing needed for an app this thin would be a minor scaling runbook that allows you to modify the replicas for the app and tune its cpu/memory requests.  You could probably copypasta strapi's runbook for for that

tier/plural/notes.tpl
```diff
@@ -1 +1,5 @@
 Your tier installation is available at https://{{ .Values.hostname }}
+
+You can inspect the runtime configuration by running:
+
+  curl -u "{{ .Values.tier.basicAuth }}:" https://{{ .Values.hostname }}/v1/whoami
```
michaeljguarino, line 5: isn't the basic auth here: `-u "auth:{{ .Values.tier.basicAuth }}"` or am I misunderstanding

tier/helm/tier/templates/deployment.yaml
gordalina: This endpoint does talk to the Stripe API, but it is a true liveness check. If it doesn't work its means its improperly configured.

tier/plural/notes.tpl
gordalina: In your example, `auth` is the user and `basicAuth` would be the password, but I'm using `basicAuth` as the user without a password. It's a common use case when you only want to simulate an API key.

gordalina:
@michaeljguarino added the scaling runbook.

michaeljguarino (approved):
Awesome, thanks for the contribution!

PR #606, opened 2023-03-30
Onboard Grafana Mimir, Grafana Agent and a multi-tenant auth solution

DavidSpek:
## Summary
Onboard Mimir for AWS and Azure initially. Onboard Grafana Agent to ship metrics to Mimir and logs to Loki. Onboard TraceShield as the auth and management layer for multi-tenant (or cluster) observability.

## Test Plan
Lot of deployments through local linking.

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [ ]  GCP


trace-shield/helm/trace-shield/values.yaml.tpl
```diff
@@ -0,0 +1,69 @@
+{{ $hydraPostgresPass := dedupe . "trace-shield.postgresHydra.password" (randAlphaNum 32) }}
+postgresHydra:
```
michaeljguarino, line 2: couldn't we have all three of these share the same db?

michaeljguarino: (you'd need to have the postgres operator create multiple dbs w/in the same one of course)

michaeljguarino (approved):
we should still rename trace shield i believe but otherwise seems fine

PR #607, opened 2023-03-30
SFTPGo: fix name, add a link to the website

drakkan:
## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


sftpgo/repository.yaml
```diff
@@ -1,9 +1,9 @@
-name: sftpgo
+name: SFTPGo
```
michaeljguarino, line 1: this needs to be downcased because it's also used as a kubernetes namespace

michaeljguarino (changes requested):
Thanks for the contribution!  One minor thing that is a technical blocker, but hope we can direct some donations your way

michaeljguarino (approved):

PR #608, opened 2023-03-31
feat: add typesense to repository

gordalina:
## Summary

This is an initial PR to get typesense into the repository so I can start testing locally. It's mostly a copy from tier. It is configured in a HA mode.

Questions:
1. I want to confirm that the persistent storage is implemented adequately? I re-used the template from `touca`.
2. I see that apps report `appVersion` in `Chart.yaml` with a `v` prefix only 32% of the time, what's the standard here? Prefix `v` or not, or follow the original app tag format?

## Test Plan

Can't test before this is merged in.

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


michaeljguarino (commented):
We just follow the semver conventions of the upstream, so either w/ or w/o the `v` prefix is fine.  Does typesense have an existing helm chart or kubernetes operator?  This is definitely the sort of app i'd expect us to want to follow some prior art for

gordalina:
> We just follow the semver conventions of the upstream, so either w/ or w/o the `v` prefix is fine. Does typesense have an existing helm chart or kubernetes operator? This is definitely the sort of app i'd expect us to want to follow some prior art for

Ok, I was using the upstream version.

I [found a helm chart](https://github.com/Spittal/typesense-helm) but I didn't use it because:
1) it hasn't been updated in 3 years
2) doesn't have a high-availability configuration

michaeljguarino (approved):
approve to unblock

PR #614, opened 2023-04-03
Jenkins fix - Marking repository public

akash2504:
## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


akash2504:
@michaeljguarino help with review.

akash2504:
Added fix for "https://github.com/pluralsh/plural-artifacts/actions/runs/4597915538/jobs/8121263624"

michaeljguarino (commented):
I think you should put `private: false` for good measure, not 100% sure the go yaml parser will behave

floreks:
I see that jenkins has a lot of images available. We might want to restrict vendor image regex a bit here not to vendor all of them. We do something similar in i.e. [prefect](https://github.com/pluralsh/plural-artifacts/blob/main/prefect/vendor_images.yaml).

floreks:
I see that jenkins has a lot of images available. We might want to restrict vendor image regex a bit here not to vendor all of them. We do something similar in i.e. [prefect](https://github.com/pluralsh/plural-artifacts/blob/main/prefect/vendor_images.yaml).

akash2504:
@michaeljguarino @floreks Addressed the review comments.

jenkins/vendor_images.yaml
```diff
@@ -1,3 +1,3 @@
 docker.io:
   images-by-tag-regex:
-    jenkins/jenkins: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
\ No newline at end of file
+    jenkins/jenkins: ^v?2\.[0-9]+?$
```
michaeljguarino, line 3: I think this will not properly vendor patch versions right?

akash2504: @michaeljguarino 
This will match 2.397 or 2.387. Check at https://regex101.com/

https://hub.docker.com/layers/jenkins/jenkins/2.397/images/sha256-4875abf22c4e465d723f5cb742218807f55d7772157ed62466c5cb45c0f9f755?context=explore

floreks: From what I can see they do not have images with the patch version defined. Do we need to also vendor all the `2.XXX-jdk` images? They provide a lot of custom tags in addition to raw `2.XXX` tags.

DavidSpek: They do have semvers for patches. See [2.387.2(https://hub.docker.com/layers/jenkins/jenkins/2.387.2/images/sha256-8f7043722b3bb576fde60fa4ab59465a4b77e677c92774514897301ab77825a3?context=explore)

jenkins/vendor_images.yaml
```diff
@@ -1,3 +1,3 @@
 docker.io:
   images-by-tag-regex:
-    jenkins/jenkins: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
```
DavidSpek, line 3: This regex isn't good. Please change it to `(?:^v?2\.(?:38[7-9]|39[0-9]|[4-9][0-9]{2}|[1-9][0-9]{3,})+?\.[0-9]+$)|(?:^v?(?:(?:[^0-2]{1})?(?:[3-9]{1}|[0-9]{2}))+\.[0-9]+?\.[0-9]+$)`. This will match anything >= v2.387.0.

DavidSpek (changes requested):
The regex needs to be changed to only include the version we care about and newer versions.

jenkins/vendor_images.yaml
DavidSpek: I've already pushed the change so you can remove your change to the image vendoring completely.

DavidSpek (approved):
LGTM

PR #617, opened 2023-04-05
Fix helm lint for typesense and redpanda.

gordalina:
## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


redpanda/helm/redpanda/templates/tests/test-connection.yaml
```diff
@@ -11,5 +11,5 @@ spec:
     - name: wget
       image: busybox
       command: ['wget']
-      args: ['{{ include "redpanda.fullname" . }}:{{ .Values.service.port }}']
+      args: ['{{ include "redpanda.fullname" . }}:{{ .Values.redpanda.listeners.http.port }}']
```
michaeljguarino, line 14: you could just rm this whole file honestly

redpanda/plural/notes.tpl
```diff
@@ -3,5 +3,5 @@ You can access the console at https://{{ .Values.consoleHostname }}
 
 Your initial admin credentials for logging into the console are:
 
-username: {{ .redpanda.secret.rootUser }}
-password: {{ .redpanda.secret.rootPassword }}
+username: {{ .Values.secret.rootUser }}
```
michaeljguarino, line 6: think these should still by `redpanda.secret.*`

DavidSpek (changes requested):
You'll need to bump the chart versions before this can be merged.

michaeljguarino (approved):
I actually think these were never uploaded, so should be good to merge as-is actually

DavidSpek (approved):
LGTM

PR #618, opened 2023-04-05
feat(jupyterhub): configure application

gordalina:
## Summary

This PR configured jupyterhub. I'll submit a follow up PR that uses non-docker images as this setups the vendor_images correctly.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


jupyterhub/vendor_images.yaml
```diff
@@ -1,3 +1,6 @@
 docker.io:
   images-by-tag-regex:
-    bitnami/jupyterhub: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
+    bitnami/jupyterhub: ^v?[0-9]+\.[0-9]+?\.[0-9]+(.*)?$
+    bitnami/jupyter-base-notebook: ^v?[0-9]+\.[0-9]+?\.[0-9]+(.*)?$
+    bitnami/configurable-http-proxy: ^v?[0-9]+\.[0-9]+?\.[0-9]+(.*)?$
+    bitnami/bitnami-shell: ^\d+(-.*)?$
```
DavidSpek, line 6: We'll need to limit these regexes so they only include the current latest version we are deploying and any newer versions. You can look in the bootstrap app for examples of how to create a regex that start at a particular semver.

DavidSpek (changes requested):
We'll need to limit the version of images being vendored a lot to prevent slow CI times and high costs.

DavidSpek (approved):
LGTM

PR #622, opened 2023-04-05
fix: add grafana agent as dependency + fix some templates

DavidSpek:
## Summary
Adds grafana-agent as a dependency for trace shield and fixes some other application templates.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/monitoring/helm/monitoring/values.yaml.tpl
+++ b/monitoring/helm/monitoring/values.yaml.tpl
@@ -6,12 +6,12 @@ global:
 kube-prometheus-stack:
   grafana:
     namespaceOverride: {{ $monitoringNamespace }}
-{{- if dig "Configuration" "grafana-agent" . }}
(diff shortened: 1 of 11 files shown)


monitoring/helm/monitoring/values.yaml.tpl
```diff
@@ -6,12 +6,12 @@ global:
 kube-prometheus-stack:
   grafana:
     namespaceOverride: {{ $monitoringNamespace }}
-{{- if dig "Configuration" "grafana-agent" . }}
+{{- if and .Configuration (index .Configuration "grafana-agent") }}
```
michaeljguarino, line 9: perhaps save as a var to reference in both cases more naturally?

DavidSpek: Done

michaeljguarino (approved):

PR #630, opened 2023-04-07
fix(dagster-gcp): fix dagster config on gcp

Walkoss:
## Summary
- Fixed GCSComputeLogManager 
- Fixed gcp service account
- Fixed dagster bucket location
- Fixed user-code-example container image for gcp
- Remove dagster-aws-env for gcp

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [x]  GCP

### DIFF
--- a/dagster/terraform/gcp/main.tf
+++ b/dagster/terraform/gcp/main.tf
@@ -26,25 +26,6 @@ module "gcs_buckets" {
   project_id            = var.project_id
   bucket_names          = [var.dagster_bucket]
   service_account_email = module.dagster-workload-identity.gcp_service_account_email
+  location              = var.bucket_location
 }
 
-resource "google_storage_hmac_key" "dagster" {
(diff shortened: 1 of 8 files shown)


dagster/terraform/gcp/main.tf
```diff
@@ -26,25 +26,6 @@ module "gcs_buckets" {
   project_id            = var.project_id
   bucket_names          = [var.dagster_bucket]
   service_account_email = module.dagster-workload-identity.gcp_service_account_email
+  location              = var.bucket_location
```
Walkoss, line 29: that might remove previously created cluster (in US by default), do we consider it as breaking change? cc @michaeljguarino

michaeljguarino: it definitely could replace buckets in place, I suppose we just need to do it though, so let's just push it

michaeljguarino (approved):

Issue #536, opened 2023-02-27
Onboard Directus.io


qorbani:
## Use Case

[Directus](https://directus.io/) is an open-source real-time API and App dashboard for managing SQL database content. It installs on top of any SQL database while you remain in complete control of your new or existing database. No need to migrate to a proprietary data model, so there's no vendor lock-in.

## Ideas of Implementation
There is already a community-driven [helm chart](https://github.com/directus-community/helm-chart) that can be used as a starting point. 

## Additional Info
[GitHub](https://github.com/directus/directus)
[Authentication](https://docs.directus.io/self-hosted/config-options.html#authentication)

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

michaeljguarino:
@qorbani thanks for the suggestion, we'll get it in line to add to the platform as soon as possible

qorbani:
Thank you @michaeljguarino! 🙏 I'm a big fan of Directus and used it in many projects. If you guys can adopt it with your ODIC, that would be HUGE! Nevertheless, having it even without that would help many people.

michaeljguarino:
@qorbani directus does work with plural oidc currently (we're also using it in prod on our new marketing website)

PR #639, opened 2023-04-13
feat: onboard directus

Walkoss:
## Summary
- Onboard directus https://github.com/pluralsh/plural-artifacts/issues/536

## Test Plan
- Deployed helm chart + terraform locally

## Next
- Set docker image from plural registry
- Configure file storage backend (s3, gcs, blob)
- Configure OIDC

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


michaeljguarino:
We should definitely use the postgres operator to set up a proper db here as well, you should be able to see a lot of prior art using the `module-library` postgres chart that wraps a lot of the helpful resources together for it

michaeljguarino:
you can see some postgres related templating here: https://github.com/pluralsh/plural-artifacts/pull/635/files

jupyterhub was weird because we had to feed it the entire postgres url, which isn't needed for directus, looks like you can just wire up these env vars: https://docs.directus.io/self-hosted/config-options.html to give it the right db configuration.  You'll almost certainly need to add `DB_SSL__REJECT_UNAUTHORIZED=false` too as an env var

it looks like directus does require s3/gcs/etc so might want to revert the `kube` terraform module and go back to the standard cloud specific ones too

PR #648, opened 2023-04-15
feat(directus): oidc + default admin account + gcs backend

Walkoss:
## Summary
- Added OIDC (OIDC users should be manually added to a role using the admin user)
- Added default admin user
- Added GCS storage backend (need to add AWS and Azure later)
- Fixed postgres connection

## Test Plan
Local linking

## Additional info
Users who log in via OIDC are not automatically admin. They have to be added to a specific role by the admin. We can set the default role of the OIDC users but not knowing the uuid of the admin role (and not being able to set it manually), it is complicated to make the OIDC users be admin by default.
We can get the Admin role uuid from the api but how to inject the `AUTH_<PROVIDER>_DEFAULT_ROLE_ID` environment variable? Any suggestion @michaeljguarino 

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [x]  GCP

### DIFF
--- a/directus/helm/directus/Chart.yaml
+++ b/directus/helm/directus/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: directus
 description: helm chart for directus
 type: application
-version: 0.1.1
+version: 0.1.2
 appVersion: 9.25.1
 dependencies:
   - name: postgres
--- a/directus/helm/directus/deps.yaml
+++ b/directus/helm/directus/deps.yaml
@@ -13,6 +13,10 @@ spec:
     name: ingress-nginx
     repo: ingress-nginx
     version: ">= 0.1.2"
+  - type: helm
+    name: postgres
+    repo: postgres
+    version: ">= 0.1.6"
   - type: terraform
     name: aws
     repo: directus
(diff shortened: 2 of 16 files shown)


michaeljguarino:
Can you point me to the docs for directus admin role assignment?  It might be pretty tricky to do this

michaeljguarino:
It's probably ok to just print out the admin user login credentials in `notes.tpl` and explain how to add users as admins after login if that's doable

Walkoss:
> Can you point me to the docs for directus admin role assignment?  It might be pretty tricky to do this

Step 7 on https://docs.directus.io/self-hosted/sso.html#google to get an example

michaeljguarino:
Yea so the best solution here is to fix an admin email and admin password, and then just document how to do the default role if necessary, but realistically you probably wouldn't want all plural logins to get admin anyway and the initial admin should assign a few manually.  It looks like a few env vars are all that's needed to set admin email and password and you could have them be recipe inputs also: https://docs.directus.io/self-hosted/quickstart.html

Walkoss:
> Yea so the best solution here is to fix an admin email and admin password, and then just document how to do the default role if necessary, but realistically you probably wouldn't want all plural logins to get admin anyway and the initial admin should assign a few manually. It looks like a few env vars are all that's needed to set admin email and password and you could have them be recipe inputs also: https://docs.directus.io/self-hosted/quickstart.html

Yes already did that. I asked for the initial admin email in the recipe configuration and then output the generated password in notes.tpl

michaeljguarino (approved):

Issue #294, opened 2022-06-14
Onboard Unleash


avaidyanatha:
We should onboard [Unleash](https://github.com/Unleash/unleash).

It's a feature flag/toggle tool that is open-source and is very flexible. The Plural marketplace currently doesn't have any feature flag  applications so this would be a great addition.

It seems like the k8s deployment is doable, here's an article: https://medium.com/@manabie/self-hosting-unleash-with-kubernetes-d6620d0dee11

walkoss:
This should be closed, Unleash is onboarded

PR #650, opened 2023-04-15
feat: onboard unleash

Walkoss:
## Summary
- Onboard https://github.com/pluralsh/plural-artifacts/issues/294
- Added OIDC

## Test Plan

- Deployed helm chart + terraform locally

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


michaeljguarino (approved):

michaeljguarino:
I suspect using the oauth proxy might also require oauth for the feature flag apis and some workaround will be needed there, probably w/ adding a dedicated ingress path for them

Walkoss:
> I suspect using the oauth proxy might also require oauth for the feature flag apis and some workaround will be needed there, probably w/ adding a dedicated ingress path for them

Good catch, I'll fix it
Should we always have OIDC support? Unleash has a auth system already

michaeljguarino:
I don't think it's necessary to always enable oidc, what you can do is have a dedicated path for the flag apis that are exposed publicly that routes to a service that is directed to the regular unleash container ports instead of the oauth proxy ones

PR #660, opened 2023-04-19
feat(weaviate): add auth + fix weaviate terraform modules

Walkoss:
## Summary
- Add authentification layer (api key)
- fix weaviate service account by rerouting recipes to correct terraform modules

## Test Plan

- Local tf link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/dagster/helm/dagster/Chart.yaml
+++ b/dagster/helm/dagster/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: dagster
 description: A Helm chart for Kubernetes
 type: application
-version: 0.1.33
+version: 0.1.34
 appVersion: 1.2.7
 dependencies:
 - name: postgres
--- a/weaviate/helm/weaviate/Chart.yaml
+++ b/weaviate/helm/weaviate/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: weaviate
 description: helm chart for weaviate
 type: application
-version: 0.1.1
+version: 0.1.2
 appVersion: 1.18.0
 dependencies:
 - name: weaviate
--- a/weaviate/helm/weaviate/values.yaml.tpl
+++ b/weaviate/helm/weaviate/values.yaml.tpl
@@ -34,11 +34,11 @@ weaviate:
     adminlist:
       users:
         - {{ .Values.adminEmail }}
+  {{ if $isGcp }}
   backups:
-    {{ if $isGcp }}
     gcs:
       enabled: true
       envconfig:
         BACKUP_GCS_BUCKET: {{ .Values.weaviateBucket }}
         BACKUP_GCS_USE_AUTH: "true"
-    {{ end }}
+  {{ end }}
--- a/weaviate/helm/weaviate/values.yaml.tpl
+++ b/weaviate/helm/weaviate/values.yaml.tpl
@@ -1,5 +1,7 @@
 {{ $isGcp := or (eq .Provider "google") (eq .Provider "gcp") }}
 
+{{ $key := dedupe . "weaviate.weaviate.admin.key" (randAlphaNum 20) }}
+
 global:
   application:
     links:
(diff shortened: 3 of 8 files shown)


dagster/helm/dagster/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: dagster
 description: A Helm chart for Kubernetes
 type: application
-version: 0.1.33
```
michaeljguarino, line 5: why is this version bumped?

weaviate/helm/weaviate/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: weaviate
 description: helm chart for weaviate
 type: application
-version: 0.1.1
```
michaeljguarino, line 5: do you need to modify the chart deps.yaml here?

dagster/helm/dagster/Chart.yaml
Walkoss: Because I upgrade dagster version in my previous PR

weaviate/helm/weaviate/Chart.yaml
Walkoss: No all good

michaeljguarino (approved):

Walkoss:
@michaeljguarino added api key authentication

weaviate/plural/notes.tpl
```diff
@@ -1 +1,5 @@
 Your Weaviate installation is available at https://{{ .Values.hostname }}
+
+Authentication by API-KEY is enabled.
+API-KEY for admin {{ .Values.adminEmail }} is {{ .weaviate.weaviate.admin.key }}
+Make sure to include this header while requesting your API -> Authorization: Bearer {{ .weaviate.weaviate.admin.key }}
```
michaeljguarino, line 5: i don't see where `.weaviate.weaviate.admin.key` is set

weaviate/helm/weaviate/values.yaml.tpl
```diff
@@ -12,6 +20,20 @@ ingress:
...
+      allowed_keys:
+        - {{ $key }}
+      users:
+        - {{ .Values.adminEmail }}
+  authorization:
+    enabled: true
+    adminlist:
+      users:
+        - {{ .Values.adminEmail }}
   backups:
```
michaeljguarino, line 37: i'm pretty sure this will have issues outside gcp since it'll set that entire map to null if you template it as

```
backups:
```

weaviate/plural/notes.tpl
Walkoss: It is set here https://github.com/pluralsh/plural-artifacts/pull/660/files#diff-8ac5632e8e4e6a4f2fd4149769678251e1a889a14ab6a0337598707f48b03ea6R16
And updated dynamically with .tpl file

weaviate/helm/weaviate/values.yaml.tpl
Walkoss: Good catch, moved condition

michaeljguarino (approved):

weaviate/helm/weaviate/values.yaml.tpl
```diff
@@ -1,5 +1,13 @@
 {{ $isGcp := or (eq .Provider "google") (eq .Provider "gcp") }}
 
+{{ $key := dedupe . "weaviate.weaviate.admin.key" (randAlphaNum 20) }}
```
michaeljguarino, line 3: so this value needs to be set here for dedupe to work properly

michaeljguarino (changes requested):

weaviate/helm/weaviate/values.yaml.tpl
Walkoss: I think dedupe will work because we wrap the chart twice, cf postgres.password dedupe used on other apps. What do you suggest?

Walkoss: Ok oh I understand, my bad

michaeljguarino (approved):

PR #661, opened 2023-04-20
fix(dagster-gcp): fix dagster user code docker image

Walkoss:
## Summary
- Fix dagster dockerfile image

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/dagster/helm/dagster/values.yaml
+++ b/dagster/helm/dagster/values.yaml
@@ -51,7 +51,7 @@ dagster:
   pipelineRun:
     image:
       repository: dkr.plural.sh/dagster/dagster/user-code-example
-      tag: 1.2.7
+      tag: 1.2.7-extended
 
   ingress:
     enabled: true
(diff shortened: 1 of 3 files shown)


dagster/helm/dagster/values.yaml
```diff
@@ -51,7 +51,7 @@ dagster:
   pipelineRun:
     image:
       repository: dkr.plural.sh/dagster/dagster/user-code-example
-      tag: 1.2.7
+      tag: 1.2.7-extended
```
michaeljguarino, line 54: do we need this pipeline run resource actually?

michaeljguarino (commented):
will need a chart vsn bump eventually

dagster/helm/dagster/values.yaml
Walkoss: No we don't

michaeljguarino (approved):

PR #663, opened 2023-04-20
fix(dagster): fix dagster dkr image

Walkoss:
## Summary
- Fix dagter image, hopefully final fix

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,7 +29,7 @@ jobs:
     - run: |
         docker build -f dagster/Dockerfile -t gcr.io/pluralsh/dagster/user-code-example:1.2.7-extended -t dkr.plural.sh/dagster/dagster/user-code-example:1.2.7-extended .
         docker push gcr.io/pluralsh/dagster/user-code-example:1.2.7-extended
-        docker push dkr.plural.sh/dagster/user-code-example:1.2.7-extended
+        docker push dkr.plural.sh/dagster/dagster/user-code-example:1.2.7-extended
   docker-sentry:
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     - run: |
         docker build -f dagster/Dockerfile -t gcr.io/pluralsh/dagster/user-code-example:1.2.7-extended -t dkr.plural.sh/dagster/dagster/user-code-example:1.2.7-extended .
         docker push gcr.io/pluralsh/dagster/user-code-example:1.2.7-extended
-        docker push dkr.plural.sh/dagster/user-code-example:1.2.7-extended
+        docker push dkr.plural.sh/dagster/dagster/user-code-example:1.2.7-extended
```
michaeljguarino, line 32: no let's leave that w/o the additional dagster

michaeljguarino: if we need to remove the double-dagster in the chart that's fine

Walkoss: Ok but it is to follow convention we have. Airflow and superset are prefix with airflow and superset respectively. Sentry is not because it's an official image I assume

michaeljguarino: actually it might be fine, let's just see.  The reason is it might cause excessive dagster deployments as that github action is rerun but it might swallow if the image already exists

michaeljguarino (approved):

PR #677, opened 2023-05-02
feat(airflow): upgrade airflow to 2.6.0

DavidSpek:
## Summary
This PR upgrade airflow to use the latest version (2.6.0) using our new automated image building pipeline so we can keep it up to date more easily. It also allows for users to use Plural deployed Redis rather than the chart internal redis.

## Test Plan
Local linking. Upgrading on existing cluster and deploying a new cluster to check the upgrade.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


rauerhans (approved):
lgtm

PR #692, opened 2023-05-19
feat: sentry update

DavidSpek:
Fixes: https://github.com/pluralsh/plural-artifacts/issues/742

## Summary

- add support for GCP and Azure
- use external clickhouse through operator-controlled clickhouse clusters
- investigate auth errors on clickhouse operator, fixes itself after some time, it's an inconvenience, but not critical
- stabilize helm install process through post-install hooks, that make sure the CHI cluster is ready to serve
- stabilize helm uninstall process through pre-delete hooks, that make sure the operator has time to wind down the CHI, before everything else gets uninstalled
- template BROKER_URL env var for rabbitmq vhost, on all sentry deployments, so it's not necessary to use the mq root credentials
- resolved DNS-resolution issues related to different behaviour of kube-dns on GCP, vs CoreDNS on AWS/Azure, where for the former pod dns names are not created unless it's a headless service, while they are always created, even ClusterIP, for the latter
- add OIDC setup docs
- use existing secrets for clickhouse creds
- bump versions across the board, as far as clickhouse compatbility allows
- narrow down vendor_images.yaml
- IMPORTANT FINDING: clickhouse publishes unpolished images under production release tags! not identifiable with RC tags, they look like they're production ready, but they're not, yet we vendor them automatically, since we have know way of knowing. Have to make sure on any updates that there's an actual github release for the image, if you bump the clickhouse-server version.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


sentry/vendor_images.yaml
```diff
@@ -1,7 +1,7 @@
...
-    getsentry/relay: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
-    getsentry/sentry: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
-    getsentry/snuba: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
-    getsentry/symbolicator: ^v?[0-9]+\.[0-9]+?\.[0-9]+$
-    yandex/clickhouse-server: ^v?[0-9]{2}\.[0-9]+?\.[0-9]+$
+    getsentry/relay: (?:^2[3-9]\.[0-9]+?\.[0-9]+$)|(?:^[3-9][0-9]\.[0-9]+?\.[0-9]+$)
+    getsentry/sentry: (?:^2[3-9]\.[0-9]+?\.[0-9]+$)|(?:^[3-9][0-9]\.[0-9]+?\.[0-9]+$)
+    getsentry/snuba: (?:^2[3-9]\.[0-9]+?\.[0-9]+$)|(?:^[3-9][0-9]\.[0-9]+?\.[0-9]+$)
+    getsentry/symbolicator: (?:^2[3-9]\.[0-9]+?\.[0-9]+$)|(?:^[3-9][0-9]\.[0-9]+?\.[0-9]+$)
+    yandex/clickhouse-server: (?:^2[2-9]\.[0-9]+?\.[0-9]+$)
```
DavidSpek, line 7: You can remove clickhouse from the image vendoring since those are handled by the operator application now.

rauerhans (approved):
lgtm, I will add the runbooks in another PR

PR #694, opened 2023-05-22
feat: add recipes for capi testing

DavidSpek:
## Summary
This PR adds the needed recipes for https://github.com/pluralsh/console/pull/357 so that the console can be tested during the Cluster API development.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


etcd/plural/recipes/etcd-azure-capi.yaml
```diff
@@ -0,0 +1,15 @@
+name: etcd-azure-capi
+description: Installs etcd on an azure aks cluster
+provider: AZURE
+primary: true
```
floreks, line 4: I think we should set it to `false` and maybe add `private: true`?

floreks (approved):

PR #697, opened 2023-05-25
feat(yatai): onbaord yatai for gcp and azure

rauerhans:
## Summary

This has both the GCP, and Azure artifacts.
- Basically there are 4 different options now.
  - AWS without ECR
  - AWS with ECR
  - Azure with Minio for S3 compatibility
  - GCP with GCS
- added runbooks
- initial setup is handled by a initialization job with a helm hook trigger

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204538190620900


yatai/helm/yatai/values.yaml
```diff
@@ -58,3 +58,8 @@ yatai-image-builder:
     serviceAccountName: "yatai"
   dockerRegistry:
     secure: true
+
```
DavidSpek, line 61: We should add default memory requests and limits and a default CPU requests for the yatai components.

DavidSpek (changes requested):
Overall good, just want to set some default resources so we don't have node instability.

DavidSpek (approved):
LGTM

PR #699, opened 2023-05-26
feat(kafka): Upgrade strimzi operator and Kafka to latest versions

DavidSpek:
## Summary
This PR updates the Strimzi Kafka Operator to the latest version and as a requirement also starts using the latest version of Kafka. Along with that the image and chart for the schema registry (needed by DataHub) was also updated.

## Test Plan
Deployed with a local link. Checked that DataHub still works.

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


DavidSpek:
@michaeljguarino That is the big question. However, the operator always only supports a limited range of Kafka versions and we are running out of time for doing this upgrade due to K8s incompatibilities. A staged upgrade would in theory be possible, but I don't think it would help much since we'd still need to upgrade to Kafka v3.

michaeljguarino:
We should probably deploy it into the prod cluster and see what breaks there, I know sentry uses kafka, think posthog does as well.  All of them are in that cluster so will be a good way to regression test

DavidSpek:
@michaeljguarino Deploy it in the prod cluster and PostHog and Sentry are running as expected.

PR #705, opened 2023-05-31
Prefect Agent Application

jimid27:
## Summary

- PR creates a Plural resource for a [prefect worker](https://github.com/PrefectHQ/prefect-helm/tree/main/charts/prefect-worker) and includes some configuration for the agent as well.  Very similar to the prefect-agent installation. 
- New feature (application)

I am unsure of how to push an image to the plural docker registry and may need assistance

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


prefect-worker/repository.yaml
```diff
@@ -0,0 +1,10 @@
+name: prefect-worker
+description: prefect-worker deployed on plural
+category: DATA
+private: true
+icon: plural/icons/prefect-worker.png
+homepage: https://www.prefect.io/
+gitUrl: https://github.com/PrefectHQ/prefect
+notes: plural/notes.tpl
+tags:
+- tag: data
```
michaeljguarino, line 10: I think this is documented in the README.md but you can add your user as a `contributor` here and that can keep the repo private but still allow you to install it for testing.

michaeljguarino:
What's the difference between this and prefect agent?  Just wondering if theoretically we could consolidate the listings

michaeljguarino (commented):
you should need these, are the git encryption filters potentially confusing you?

jimid27:
> What's the difference between this and prefect agent? Just wondering if theoretically we could consolidate the listings

Good question - Workers and Agents are very similar to one another but workers offer a lot more control over infrastructure configuration and have the ability to route work to specific types of execution environments

jimid27:
> > > you should need these, are the git encryption filters potentially confusing you?
> > 
> > 
> > Potentially haha - Kept seeing binary file which I hadn't seen before but likely just that
> 
> You wouldn't want to check in the binary files here, they are only there when generated in an installation repo for security (if you look at the charts for other apps you'll see they aren't encrypted)

Ok I think we are good to go on the values files now 🤞

michaeljguarino (approved):

PR #707, opened 2023-06-01
feat(elasticsearch): Upgrade/elastic

rauerhans:
## Summary

- Upgrade helm chart  from `2.4.0` to `2.8.0`
- Upgrade ECK operator from `2.4.0` to `2.8.0`
- Upgrade supported ELK app versions from `7.13.2` to `7.17.0`
  - newest version is actually `8.8.0`, however there's an upgrade path that foresees a stepwise approach that requires an upgrade to specifically `7.17.0` first. We should let it sit at `7.17.0` for a while and then upgrade from there at a later date, otherwise we might break people's running ELK clusters. See also: https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html
- vendoring operator image now
- not vendoring the ELK stack images, there are too many of them

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


elasticsearch/helm/elasticsearch/values.yaml
```diff
@@ -39,18 +43,21 @@ elasticsearch:
       memory: 4Gi
       cpu: 2
   storage: 50Gi
-  version: '7.13.2'
+  version: '7.17.0'
```
DavidSpek, line 46: Why is this a string while the others aren't?

DavidSpek (changes requested):
Can you setup the CRDs so that they use the vendored images? All `7.17.0` images should be available now from our registry.

michaeljguarino (approved):

michaeljguarino (approved):

rauerhans:
> Can you setup the CRDs so that they use the vendored images? All `7.17.0` images should be available now from our registry. this is done

DavidSpek (approved):
LGTM

PR #708, opened 2023-06-01
fix: prefect icon ref

jimid27:
## Summary

- Fix for the icon name which was causing the vendor push to fail

## Test Plan

N/A

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


michaeljguarino (approved):

DavidSpek (changes requested):
Can you please bump the helm chart vesion so it can be pushed properly,

DavidSpek:
Closing as this seems to be handled in https://github.com/pluralsh/plural-artifacts/pull/709 already.

PR #712, opened 2023-06-02
feat(mlflow): update and onboard on GCP + Azure

DavidSpek:
## Summary
Still untest MLFLow upgrade.

This should be good to go now.

- upgraded to latest mllfow version `2.5.0`
- added support for GCP and Azure
  - gcp with gs
  - azure with blob storage container (needs credentials in any training job as laid out in example repo)

tested on all three with this example repo: https://github.com/pluralsh/mlflow-on-k8s-example

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino (commented):
How smooth is the process without waiting for migrations in the mlflow server?  We might want to at least wait for postgres to be available with `nc`

michaeljguarino:
Another thing with mlflow @rauerhans could you add support for the version tagging we do in airbyte and some other apps so we can create versioned release channels?  It just requires including the test base helm chart and some minor config.

rauerhans:
Made some ammendments:
- added an initContainer to the deployment that `until nz -zv` checks and waits until the postgres is up (it was fairly smooth before, mlflow waits and retries with the db migration, this is definitely cleaner tho as it completes on the first try)
- added the `test-base` subchart and configuration mirroring other apps where we use it
- added the runbook (missed that previously, my bad)

michaeljguarino (approved):
awesome

PR #728, opened 2023-06-02
ci: update renovate config

DavidSpek:
This PR updates the renovate config so it will group the updates for new applications. It also adds the script used to update the renovate config so it can be run more easily.


rauerhans (approved):
lgtm..

PR #729, opened 2023-06-02
chore: Update prefect worker inputs & update prefect versions

jimid27:
## Summary

- Bump Prefect image ref in chart (and subsequently bump chart)
- Add workPool string input (populates: https://github.com/PrefectHQ/prefect-helm/blob/main/charts/prefect-worker/values.yaml#L39)

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP

### DIFF
--- a/prefect-worker/plural/recipes/prefect-worker-aws.yaml
+++ b/prefect-worker/plural/recipes/prefect-worker-aws.yaml
@@ -19,6 +19,9 @@ sections:
   - name: workspaceId
     documentation: the workspace id of your Prefect workspace
     type: STRING
+  - name: workPool
+    documentation: the work pool associated with the prefect worker
+    type: STRING
   items:
   - type: TERRAFORM
     name: aws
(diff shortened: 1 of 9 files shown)


prefect-worker/plural/recipes/prefect-worker-aws.yaml
```diff
@@ -19,6 +19,9 @@ sections:
   - name: workspaceId
     documentation: the workspace id of your Prefect workspace
     type: STRING
+  - name: workPool
```
DavidSpek, line 22: Wouldn't you want to template the value you get from this recipe input into the helm chart using the `yalues.yaml.tpl` file?

DavidSpek (commented):
We can merge it, but I think something is missing @jimid27.

prefect-worker/plural/recipes/prefect-worker-aws.yaml
jimid27: hmm - I think it is [here](https://github.com/pluralsh/plural-artifacts/blob/main/prefect-worker/helm/prefect-worker/values.yaml.tpl#L3) but lemme know if there is something more I'm missing!  Was more or less following what had been done previously for the accountId and workspaceId inputs.

michaeljguarino (approved):

PR #734, opened 2023-06-06
chore: Prefect-Agent remove API key from helm ref

jimid27:
## Summary

- Currently in the `values.yml.tpl` we include the user inputted prefect api key which then is shown in plaintext in the `default-values.yaml` file after a plural build.  The value is ref'd in the secrets.yaml template and should not be needed in the helm template values. 
- Feedback is welcome if the api key needs to be populated here, but if it is expected, we would want to omit it from the git push from the customer side post build as there is an inherent security risk.

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP

### DIFF
--- a/prefect-agent/helm/prefect-agent/Chart.yaml
+++ b/prefect-agent/helm/prefect-agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prefect-agent
 description: helm chart for prefect-agent
 type: application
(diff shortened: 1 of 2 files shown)


jimid27:
> @jimid27 that file will be encrypted in git directly, so there's no security risk to be concerned about actually

oh duh good point.  closing!

PR #736, opened 2023-06-07
feat(typesense): onboard typesense

rauerhans:
## Summary

(actually) onboards typesense
- typesense is an alternative to angolia
- there's no official helm chart, or any other officially supported K8s deployment, so we had to create the chart in this artifact from scratch
- it supports HA through the raft protocol
- a HA-cluster of typesense nodes is implemented via STS
- used this script to test the API, which is exposed via ingress: https://github.com/pluralsh/typesense-test
### Problem with 3 replicas during rolling pod update:
- during the roll, where one pod, let's call it pod 1, get’s terminated, the other two (if needed) select a new leader, and form a new quorum capable cluster
- when pod 1 is restarted it needs to reach `ready` state for it to receive traffic and join the HA cluster. This takes a couple of seconds for the peering to figure itself out.
- however, at the same time, i.e. as soon as pod 1 becomes `ready`, the STS/Replica controller will already move on with the roll and terminate one of pod 2, or pod 3
- this leaves the HA cluster in a broken state, because it's almost certain that pod 1 didn't get a chance to join the cluster yet
#### Solution
- ensure the number of STS replicas is at least 5
  - as laid out above, during a roll, the raft cluster actually suffers loss of two nodes, this case is covered with 5 or more replicas
-  add a PDB of with `maxUnavailable=1` (or 2 with 8 or more replicas, or 3 with 10 or more, etc..)
  - this should make sure that no other rolling strategy but `OrderedReady` is possible
- use a high `initialDelaySeconds` in the readiness probe in the pod template of the STS, that covers the time a pod needs to rejoin the raft cluster

#### References
We should link these important gh issue conversations that helped understand the idiosyncracies of the k8s deployment of typesense:
- https://github.com/typesense/typesense/issues/203
- https://github.com/typesense/typesense/issues/465

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


DavidSpek (approved):
LGTM

michaeljguarino (approved):

PR #755, opened 2023-06-29
feat: Add directus s3 support

michaeljguarino:
## Summary
Needed to persist files durably

## Test Plan
local links

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/directus/plural/recipes/directus-azure.yaml
+++ b/directus/plural/recipes/directus-azure.yaml
@@ -19,6 +19,10 @@ sections:
     - name: hostname
       documentation: FQDN to use for your directus installation
       type: DOMAIN
+    - name: directusBucket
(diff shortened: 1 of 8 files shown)


DavidSpek (approved):
LGTM

directus/plural/recipes/directus-azure.yaml
```diff
@@ -19,6 +19,10 @@ sections:
     - name: hostname
       documentation: FQDN to use for your directus installation
       type: DOMAIN
+    - name: directusBucket
+      documentation: the bucket for blob assets stored in directus, like images
+      type: BUCKET
+      default: directus
```
walkoss, line 25: This will not be used as we don't handle blob storage for now and can confuse people running directus on azure

Issue #177, opened 2022-03-04
Onboard cube.dev


michaeljguarino:
## Use Case
cube.dev is a pretty promising analytics visualization solution that many users might find useful

## Ideas of Implementation
It looks like they don't have a working helm chart, but do have some k8s documentation already: https://cube.dev/docs/deployment/platforms/kubernetes

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

walkoss:
@ismaelheelio cube is onboarded!

ismaelheelio:
Thanks Walid! Good job!

PR #757, opened 2023-07-03
feat: onboard cube

walkoss:
## Summary

- Onboard **cube** ([github](https://github.com/cube-js/cube), [website](https://cube.dev/)), the Semantic Layer for Building Data Applications
- Fix https://github.com/pluralsh/plural-artifacts/issues/177

## Test Plan

- Private publisher with build and deploy
- Tested to query a table on postgres using API

### Install cube on a plural cluster
plural bundle install cube
plural build --only cube
plural deploy --from cube

### Wait for cube to be installed
plural watch cube

### Query example result with python sdk
Install cube package
`pip install CubeJsClient`

In a python intepreter, enter the following and replace variables
```python
from cube_js_client import CubeJsClient
client = CubeJsClient("https://<YOUR_CUBE_HOST>", "<YOUR_JWT_SECRET>")
results = client.load({
  "measures": [
    "orders.count"
  ],
  "timeDimensions": [
    {
      "dimension": "orders.created_at",
      "granularity": "day"
    }
  ],
  "order": {
    "orders.created_at": "asc"
  }
})
print(results)
```

### Destroy app
plural destroy cube

## Checklist

    - [ ]  inputing secrets
    - []  Azure
    - []  AWS
    - [x]  GCP


cube/helm/cube/values.yaml.tpl
```diff
@@ -0,0 +1,24 @@
...
+
+cube:
+  ingress:
+    hostname: {{ .Values.hostname }}
+  config:
+    apiSecret: {{ $apiSecret }}
+
+cubestore:
+  cloudStorage:
+{{ if $isGcp }}
```
michaeljguarino, line 18: this will probably fail for non-gcp becasue the yaml will look like a nil

cube/helm/cube/templates/schema.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cube-schema
+data:
+{{ (.Files.Glob "schemas/**.yaml").AsConfig | indent 2 }}
```
michaeljguarino, line 6: I think we might need to move these to the values file since users can't customize the file imports directly in the wrapper chart

walkoss: Yes, I've added documentation explaining how to add personal models. They will create another configmap which will be referenced in values.yaml
This one is just an example (same as the default user deployment for dagster)

michaeljguarino: that's probably fair actually, i probably didn't tie in the docs reading the PR

michaeljguarino (approved):

PR #758, opened 2023-07-03
feat: Mindsdb Application

smyja:
## Summary

Initial setup for Mindsdb

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


mindsdb/repository.yaml
```diff
@@ -0,0 +1,6 @@
+name: mindsdb
+description: mindsdb deployed on plural
+category: DATA
+private: true
+icon: plural/icons/mindsdb.png
+notes: plural/notes.tpl
```
michaeljguarino, line 6: you should add yourself as a contributor here

smyja: okay

mindsdb/repository.yaml
```diff
@@ -0,0 +1,12 @@
...
+category: DATA
+private: false
+releaseStatus: BETA
+icon: plural/icons/mindsdb.png
+notes: plural/notes.tpl
+homepage: https://mindsdb.com
+gitUrl: https://github.com/mindsdb/mindsdb
+
+contributors:
+- Oghenemaro Akpobi
```
michaeljguarino, line 12: this would need to be the email you use to login with plural actually

smyja: > this would need to be the email you use to login with plural actually

okay

mindsdb/helm/mindsdb/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
michaeljguarino, line 1: should probably remove this file

mindsdb/helm/mindsdb/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
+kind: Deployment
```
michaeljguarino, line 2: Isn't this a stateful application?  Deployments are usually not the right construct here, and should use a statefulset to simplify volume provisioning/binding.   Also it might be useful to search for and use an existing mindsdb helm chart instead of writing it fresh if you're a bit unfamiliar with k8s.

michaeljguarino (commented):
There's a lot missing here still, I can merge it to get it registered in our api and help unblock, but there'd definitely need to be a bit done after

mindsdb/helm/mindsdb/templates/deployment.yaml
smyja: > Isn't this a stateful application?  Deployments are usually not the right construct here, and should use a statefulset to simplify volume provisioning/binding.   Also it might be useful to search for and use an existing mindsdb helm chart instead of writing it fresh if you're a bit unfamiliar with k8s.

This is an existing mindsdb helm chart. It's on their repo, although they say it was an open source contribution from someone.

michaeljguarino: Does the server store any data underneath or is it actually effectively stateless?

smyja: it does

> Does the server store any data underneath or is it actually effectively stateless?

it's stateful, i'll cange it

mindsdb/helm/mindsdb/templates/tests/test-connection.yaml
smyja: I have removed it

smyja:
This one's problematic. I'll create for another app.

PR #763, opened 2023-07-05
ci: revert renovate config update

DavidSpek:
This reverts commit 79c4d532eb2214878d2288d91c2e0b894b7fba6d.

The version of renovate we use doesn't support the new value yet.


rauerhans (approved):
stamp

PR #769, opened 2023-07-07
feat: upgrade dagster

kwigley:
## Summary

bumps dagster to 1.3.13

## Test Plan

tested using plural on eks

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


dagster/helm/dagster/values.yaml
```diff
@@ -106,7 +106,7 @@ dagster:
       - name: "k8s-example-user-code-1"
         image:
           repository: ghcr.io/pluralsh/containers/dagster/user-code-example
-          tag: 1.3.8-plural1.0.4
+          tag: 1.3.13-plural1.0.4
```
kwigley, line 109: not sure what use use here

kwigley: ah, I see. thanks!

michaeljguarino (approved):

PR #772, opened 2023-07-11
fix(sentry): resolve errors with sentry runbooks

DavidSpek:
## Summary
Fixes a number of issues with the runbooks for Sentry. Also contains a bit of cleanup to have a logical sorting of the different components.

## Test Plan
Local linking

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


sentry/helm/sentry/templates/runbooks.yaml
rauerhans, line 1: idk, do we need one separate runbook for the symbolicator? I think it has only one component

rauerhans (approved):
lgtm

sentry/helm/sentry/templates/runbooks.yaml
DavidSpek: It has only 1 component (just like relay), but they are separate images and components within Sentry. Similar to Snuba.

PR #777, opened 2023-07-17
feat: Create Azure MSI in Terraform

maciaszczykm:
## Summary
Azure MSI will be used when clusters will be migrated to CAPZ. Adding it now to simplify migration in the future.

## Test Plan
It was tested manually. See https://github.com/pluralsh/plural-artifacts/pull/741/commits/9ba6896018b400c312ba840c47a9c904d17c8f3e.

## Checklist
    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


maciaszczykm:
Let's put it on hold, as I have just started testing it and there will be a problem with it. The problem is order of executing Helm and Terraform, Helm creates cluster now, so it should be executed first, but on the other hand it will not succeed without identity that would be created by Terraform.

PR #787, opened 2023-07-26
feat(cube): add aws s3 bucket

walkoss:
## Summary

- Add possibility to confiigure s3 bucket for aws cloud 

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/cube/helm/cube/Chart.yaml
+++ b/cube/helm/cube/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cube
 description: helm chart for cube
 type: application
-version: 0.1.8
+version: 0.1.9
 appVersion: "v0.33.30"
 dependencies:
 - name: cube
--- a/cube/terraform/aws/deps.yaml
+++ b/cube/terraform/aws/deps.yaml
@@ -2,7 +2,7 @@ apiVersion: plural.sh/v1alpha1
 kind: Dependencies
 metadata:
   description: cube aws setup
-  version: 0.1.0
+  version: 0.1.1
 spec:
   dependencies:
   - name: aws-bootstrap


cube/helm/cube/values.yaml.tpl
```diff
@@ -13,12 +13,20 @@ cube:
...
   cloudStorage:
+    {{ if $isGcp }}
     gcp:
       credentialsFromSecret:
         name: cube-gcp-credentials
         key: key.json
       bucket: {{ .Values.cubeBucket }}
+    {{ else if eq .Provider "aws" }}
+    aws:
+      accessKeyID: {{ importValue "Terraform" "iam_access_key_id" }}
```
michaeljguarino, line 27: it would be better to use IRSA instead of fixed keys, I'm not sure if cubejs supports it but might be worth looking

cube/terraform/aws/iam.tf
```diff
@@ -0,0 +1,23 @@
+module "assumable_role_cube" {
```
michaeljguarino, line 1: this would only need to be used for IRSA if it's relevant

cube/helm/cube/values.yaml.tpl
walkoss: Unfortunately, cubejs doesn't seems to support IRSA (nor workload identity)

cube/plural/recipes/cube-aws.yaml
```diff
@@ -13,6 +13,10 @@ sections:
     - name: hostname
       type: DOMAIN
       documentation: the fully qualified domain name your cube api will be available at
+    - name: cubeBucket
+      type: BUCKET
+      documentation: gcs bucket for storing cube files
```
michaeljguarino, line 18: gcs bucket -> s3 bucket

cube/terraform/aws/iam.tf
```diff
@@ -0,0 +1,23 @@
...
+  source                        = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version                       = "3.14.0"
+  create_role                   = true
+  role_name                     = "${var.cluster_name}-${var.role_name}"
+  provider_url                  = replace(data.aws_eks_cluster.cluster.identity[0].oidc[0].issuer, "https://", "")
+  role_policy_arns              = [module.s3_buckets.policy_arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:${var.namespace}:${var.cube_serviceaccount}"]
+}
+
+resource "aws_iam_user" "cube" {
```
michaeljguarino, line 11: I believe this iam user is no longer needed

michaeljguarino (approved):

michaeljguarino (changes requested):
actually wait, you need to bump the versions for both the tf and helm

PR #791, opened 2023-07-29
Add postgres user for s3 backup/restore

michaeljguarino:
## Summary

There seems to be a bug in the wal-g restore process in the postgres operator that prevents it from authing with IRSA. For now just add a fixed keypair to use in backups which will be synced into the console's new pg management backend.

## Test Plan
link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/postgres/terraform/aws/deps.yaml
+++ b/postgres/terraform/aws/deps.yaml
@@ -2,7 +2,7 @@ apiVersion: plural.sh/v1alpha1
 kind: Dependencies
 metadata:
   description: postgres aws setup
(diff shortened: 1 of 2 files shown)


rauerhans:
looks good, with these env vars you'll have to explicitly set them in the backend code, I think, because any AWS sdk usually expects AWS_ACCESS_KEY_ID (you have just ACCESS_KEY_ID)

rauerhans (approved):

PR #801, opened 2023-08-09
Support custom node launch template

michaeljguarino:
## Summary
Som users want this to disable imdsv1 but it's honestly a decent feature we should just support regardless.

## Test Plan
need to local link

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/bootstrap/plural/docs/custom-launch-template.md
+++ b/bootstrap/plural/docs/custom-launch-template.md
@@ -1,0 +1,15 @@
+# Using a Custom Launch Template in Plural AWS Provider
+
+For Plural AWS clusters, it's sometimes useful to define your own node launch template.  Some common usecases are:
+
+* toggling the setup of security settings like whether to require Instance Metadata Service v2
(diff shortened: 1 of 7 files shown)


DavidSpek:
We shouldn't merge this since this is already being done for the sysbox work.

rauerhans:
I will add that this will most likely work, but I already did that work for sysbox in the module-library repo. Leveraging that for the bootstrap artifact would require to change the source of the nodegroup module from  `github.com/pluralsh/module-library//terraform/eks-node-groups/single-az-node-groups?ref=20e64863ffc5e361045db8e6b81b9d244a55809e` to `github.com/pluralsh/module-library//terraform/eks-node-groups/multi-az-node-groups?ref=feat-ubuntu-ng`

DavidSpek:
@rauerhans Can you please create a new PR to update the bootstrap module to use the new versioning of the module and include the launch template module?

Issue #217, opened 2022-04-13
Onboard https://backstage.io/


michaeljguarino:
## Use Case
They look like one of the better kubernetes app managers out there and could be a good stopgap for our internal app dev solution.

## Ideas of Implementation
Can likely build off of https://backstage.io/docs/deployment/helm

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

PR #803, opened 2023-08-10
feat: onboard backstage

walkoss:
## Summary

Onboard Backstage ([website](https://backstage.io/), [github](https://github.com/backstage/backstage)), fixes https://github.com/pluralsh/plural-artifacts/issues/217

## Test Plan

- Local link to deploy

## Checklist

    - [ ]  inputing secrets
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP


backstage/vendor_images.yaml
```diff
@@ -0,0 +1,4 @@
+ghcr.io:
+  images:
+    backstage/backstage:
```
michaeljguarino, line 3: could we use the regex format to start vendoring future vsns too?

walkoss: Yes, fixed

michaeljguarino (approved):

PR #804, opened 2023-08-10
fix(tempo): mark recipes as public

DavidSpek:
## Summary
The recipes were still marked as private. This PR fixes that.


michaeljguarino (approved):
we might need to explicitly set `private: false` actually, it might not encode them as attributes sent to gql w/o it

PR #811, opened 2023-08-28
feat(bootstrap): upgrade AWS EKS to K8s 1.24 + NAT improvements

DavidSpek:
## Summary
This PR upgrade the Kubernetes version for our AWS EKS deployments to 1.24 since the EOL is coming up. It also takes the opportunity of this breaking change to improve the VPC and subnet configuration a bit. While working on our migration to Cluster API an issue was observed where the migration got stuck as the CAPA pod lost communication to AWS while it was shuffling around the NAT gateways and route tables of the subnets the workers are deployed in. The reason to have a NAT gateway in each subnet is to avoid cross AZ ingress/egress charges for any data going to/from the public internet. Applying this change to the networking setup as part of the Terraform should help ensure the migration to CAPI is more smooth.

## Test Plan
Local linking

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/bootstrap/terraform/aws-bootstrap/deps.yaml
+++ b/bootstrap/terraform/aws-bootstrap/deps.yaml
@@ -1,8 +1,8 @@
 apiVersion: plural.sh/v1alpha1
 kind: Dependencies
 metadata:
-  description: Creates an EKS cluster and prepares it for bootstrapping 
-  version: 0.1.51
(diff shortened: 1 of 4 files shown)


michaeljguarino (approved):
We should probably notify people of this manually as well

PR #821, opened 2023-09-04
feat(bootstrap): bump aws addon versions

DavidSpek:
## Summary
This PR bumps the addons for AWS clusters. The VPC CNI addon needs to be bumped to `1.13` before it can be bumped to the latest `1.14` version.


michaeljguarino (approved):
are we sure this doesn't need to be breaking?

PR #829, opened 2023-09-05
WIP: create dedicated app for cert-manager

davidspek:
## Summary
This PR creates a dedicated app for `cert-manager` so that its version can be maintained more easily, it's more clear to users what's installed on their cluster and what versions the apps are, to improve security by isolation cert-manager into its own namespace, and to be able to improve the new CAPI bootstrapping steps (so providers are installed after cert-manager is running). This change needs further testing since on Azure and GCP to ensure the cloud permissions work correctly, as well as with Plural DNS although that likely won't have any issues.

## Test Plan

- [x] Tested that AWS Route53 still works
- [ ] Tested that GCP Cloud DNS still works
- [ ] Tested that Azure DNS still works
- [ ] Tested when using Plural DNS
- [ ] Tested while used in the CAPI deployment and migration

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/cert-manager/terraform/gcp/output.tf
+++ b/cert-manager/terraform/gcp/output.tf
@@ -1,3 +1,3 @@
 output "service_account_email" {
-  value = module.certmanager_workload_identity.gcp_service_account_email
+  value = module.cert_manager_workload_identity.gcp_service_account_email
 }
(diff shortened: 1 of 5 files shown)


cert-manager/terraform/gcp/output.tf
```diff
@@ -0,0 +1,3 @@
+output "service_account_email" {
+  value = module.certmanager_workload_identity.gcp_service_account_email
```
floreks, line 2:
```suggestion
  value = module.cert_manager_workload_identity.gcp_service_account_email
```

PR #836, opened 2023-09-06
feat(bootstrap): add capi testing recipes + cleanup old ones

DavidSpek:
## Summary
This PR adds the recipes we are currently using to test the new CAPI deployment and migration to the main branch. It also removes old recipes that were used with a previous approach to the CAPI deployment. Another thing it does is update the recipe for Azure CAPI deployments to use the main terraform module since the changes required for the CAPI migration have been merged into the currently used terraform module already.


maciaszczykm (approved):
LGTM from Azure perspective

PR #822, opened 2023-09-05
feat: Update azure-bootstrap and azure-identity in preparation for CAPI release

maciaszczykm:
## Summary
Extract azure-bootstrap and azure-identity changes from https://github.com/pluralsh/plural-artifacts/pull/741 in preparation for CAPI release. Additionally bump azure-bootstrap `provider_vsn` and update `sku_tier` to be compatible with it.

## Test Plan
```sh
export EXP_PLURAL_CAPI=false
plural init
plural bundle install bootstrap azure-k8s
plural build
plural link terraform bootstrap --name azure-bootstrap --path ../plural-artifacts/bootstrap/terraform/azure-bootstrap
plural link helm bootstrap --name azure-identity --path ../plural-artifacts/bootstrap/helm/azure-identity
plural build

# Set following required_providers in bootstrap/terraform/main.tf. It's required as provider_vsn will not be used during local link.
#   required_providers {
#    azurerm = {
#      source = "hashicorp/azurerm"
#      version = "~> 3.70.0"
#    }
#    kubernetes = {
#      source  = "hashicorp/kubernetes"
#      version = "~> 2.23.0"
#    }
#  }

plural deploy --commit "deploy"
plural destroy
```

```sh
export EXP_PLURAL_CAPI=false
plural init
plural bundle install bootstrap azure-k8s
plural build
plural deploy

plural link terraform bootstrap --name azure-bootstrap --path ../plural-artifacts/bootstrap/terraform/azure-bootstrap
plural link helm bootstrap --name azure-identity --path ../plural-artifacts/bootstrap/helm/azure-identity
plural build

# Set following required_providers in bootstrap/terraform/main.tf. It's required as provider_vsn will not be used during local link.
#   required_providers {
#    azurerm = {
#      source = "hashicorp/azurerm"
#      version = "~> 3.70.0"
#    }
#    kubernetes = {
#      source  = "hashicorp/kubernetes"
#      version = "~> 2.23.0"
#    }
#  }

plural deploy  --commit "deploy"
plural destroy
```
## Checklist

- [ ]  No images hosted from dockerhub
- [ ]  Are dashboards present to understand the health of the application.  There **must** be at least 1 of these
    - [ ]  all databases should have dashboards
    - [ ]  ideally also have at least cpu/mem utilization dashboards for webserver tier of the app
    - [ ]  you can use `plural from-grafana` to convert a grafana dashboard found via google to our CRD
- [ ]  Are scaling runbooks present
[...]

PR #837, opened 2023-09-06
feat(bootstrap): TF changes for AWS CAPI migration

DavidSpek:
## Summary
This PR updates the AWS bootstrap terraform so that it is ready for the CAPI migration. Similar to https://github.com/pluralsh/plural-artifacts/pull/835 and https://github.com/pluralsh/plural-artifacts/pull/822. It also updates the recipe we are using to the CAPI and the migration to use the main aws bootstrap terraform.


bootstrap/terraform/aws-bootstrap/main.tf
```diff
@@ -36,6 +38,8 @@ module "vpc" {
 }
 
 module "cluster" {
+  count =  var.create_cluster ? 1:0
```
michaeljguarino, line 41: shouldn't the `create_eks` field already handle this?

bootstrap/terraform/aws-bootstrap/main.tf
```diff
@@ -3,6 +3,8 @@ data "aws_availability_zones" "available" {}
 data "aws_caller_identity" "current" {}
 
 module "vpc" {
+  count =  var.create_cluster ? 1:0
```
michaeljguarino, line 6: similar there's a create vpc field we can use to turn this module off

DavidSpek: I don't think that variable was ever implemented. If not, it's easier to consolidate them.

michaeljguarino (approved):

PR #841, opened 2023-09-08
feat: upgrade airbyte to 0.50.25

kwigley:
## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP

### DIFF
--- a/airbyte/helm/airbyte/Chart.yaml
+++ b/airbyte/helm/airbyte/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: airbyte
 description: Unified data integration platform
 type: application
-version: 0.4.25
-appVersion: 0.50.8
+version: 0.4.26
+appVersion: 0.50.25
 dependencies:
 - name: airbyte
   version: 0.43.22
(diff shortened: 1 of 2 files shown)


airbyte/helm/airbyte/Chart.lock
```diff
@@ -1,7 +1,7 @@
 dependencies:
 - name: airbyte
-  repository: file://../../../../airbyte/charts/airbyte
-  version: 0.43.22
+  repository: https://airbytehq.github.io/helm-charts
```
kwigley, line 3: @michaeljguarino if there are upstream changes in your fork that should be included, I'll just need a hand to incorporate those! Thanks

michaeljguarino: yea my fork has some meaningful changes from last i remember

michaeljguarino: could you revert that for now as we can't merge with it as a dependency (you actually shouldn't need to have my fork to perform this upgrade as the chart is in the repo and should be safe to apply w/

michaeljguarino: yep that's it

michaeljguarino (commented):
I'm seeing an issue where airbyte-webapp can't bind port 80 and so cannot run (likely the base image that they're using no longer runs as root).  I don't see an easy fix there so we might need to pause until there's a way to solve it.

kwigley:
Ah ya, I see. I tested with their most recent helm chart and everything I seemed to be working properly. Sounds good, I'll close for now.

PR #842, opened 2023-09-08
feat: Begin Onboarding temporal

michaeljguarino:
## Summary
A number of users have requested this and i believe this gives a decent starting point to test the packaging.

## Test Plan
deployed and verified on prod cluster

## Checklist

    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


Makefile
```diff
@@ -25,3 +25,8 @@ helm-dependencies-%: # syncs helm dependencies for a chart
 
 upload-%: # uploads artifacts
 	plural apply -f $*/Pluralfile
+
+download-temporal-chart: # downloads temporal's helm chart since it isn't published by them
```
DavidSpek, line 29: I think this can be removed now since we're using our helm charts repo

DavidSpek (approved):
LGTM

PR #857, opened 2023-09-27
fix(chatwoot): set correct app version

DavidSpek:
## Summary
The app version wasn't updated in the last update. This PR updates it so it doesn't confuse users.


rauerhans (commented):
lgtm

rauerhans (approved):

PR #868, opened 2023-10-06
docs: Add docs for running dbt core in airflow via cosmos

chrishronek:
## Summary
I noticed that you have a lot of dbt adapters as part of your base image, if you advised customers to follow this guide, you could avoid the pain of maintaining those.

## Test Plan
I've successfully deployed this to a plural.sh Airflow instance to confirm that it works.

## Checklist
    - [x]  ideally also have at least cpu/mem utilization dashboards for webserver tier of the app
    - [x]  you can use `plural from-grafana` to convert a grafana dashboard found via google to our CRD
    - [x]  inputing secrets
    - [x]  configuring autoscaling
    - [x]  Azure
    - [x]  AWS
    - [x]  GCP

___

Again, not sure if any of the above were really applicable since this is just a doc, but let me know if you'd like me to change anything!


michaeljguarino (commented):
could we drop the png?  I don't know if the markdown renderer will handle that properly actually.

michaeljguarino (approved):

PR #881, opened 2023-11-03
docs: Scaling horizontally on kubernetesexecutor (airflow)

chrishronek:
## Summary
Just a simple docs add to help plural users to scale their Airflow workloads when using the KubernetesExecutor

## Test Plan
These are steps that I've taken in my own Plural Airflow application and they worked wonders for me.

## Checklist
Deleted the checklist because it didn't really apply to a docs addition

If I missed anything, please let me know 🍻


michaeljguarino (commented):
is there a reason the custom node doc was removed?

chrishronek:
> is there a reason the custom node doc was removed?

I decided to combine it with this doc because it felt like they were the same topic

PR #907, opened 2024-03-29
feat(kafka): allow pod template customizations

Cbeck527:
## Summary
Allow pod template customizations on the strimzi kafka CRD. This will allow operators to set up `podAffinity` blocks and more.

A good use case for this feature is configuring all kafka/zookeper pods to run on different nodes: [strimzi docs](https://strimzi.io/docs/operators/0.35.1/full/configuring#configuring-pod-anti-affinity-to-schedule-each-kafka-broker-on-a-different-worker-node-str)

### DIFF
--- a/kafka/helm/kafka/Chart.yaml
+++ b/kafka/helm/kafka/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kafka
 description: A Helm chart for Kubernetes
 type: application
-version: 0.2.1
+version: 0.2.2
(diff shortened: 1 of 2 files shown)


michaeljguarino (commented):
You'll need to bump the version in Chart.yaml too for us to be able to publish but lgtm otherwise

Cbeck527:
bumped- thanks!

michaeljguarino (approved):

PR #920, opened 2024-10-08
feat(datahub): Bump version to 0.13.2

darrylng:
## Summary
Bumps the version of Datahub to a newer version. Not quite the latest but this version bump requires the least amount of changes in terms of dependencies to `elasticsearch` and `kafka`.

I am unsure if the changes in the `Chart.lock` should be committed, as I'm not sure if the SHA will change once plural's CICD packages the charts and duplicates the repositories. Please advise. I am also not sure when the images are pushed to dkr.plural.sh.

Additional reference here: https://artifacthub.io/packages/helm/datahub/datahub/0.4.19

## Test Plan
I tested this change using the following steps:
1. Added helm repo from datahub: https://helm.datahubproject.io
2. Point image repository to datahub's docker registry: https://hub.docker.com/u/acryldata
3. Ensure `helm dependency upgrade` works
4. Use `plural link ...` to link to my local project
5. Run `plural build --only datahub --force && plural deploy`
6. Ensure containers all run happily

## Checklist
    - [ ]  inputing secrets
    - [ ]  Azure
    - [ ]  AWS
    - [ ]  GCP


michaeljguarino (approved):
awesome thanks for testing this out.  The Chart.lock should be included, as we don't run `helm dependencies update` before publishing the chart.

